PR #5, opened 2020-08-03
Initial Pi 2 + Pi Zero support

kenbell:
Adds basic support for Pi 2 and Pi Zero.

Pi Zero also requires: https://github.com/f-secure-foundry/tamago-go/pull/2 (Pi Zero does not support the `lsb`, `isb`, `dmb` instructions, so instead does equivalents using MCR-style instructions.


kenbell:
@abarisani I didn't see any documents on the contribution process to this repo.  Is raising this PR sufficient?

abarisani:
> @abarisani I didn't see any documents on the contribution process to this repo. Is raising this PR sufficient?

Yes it is. However we are all on vacation so please have some more patience :). Thanks!

abarisani:
Some initial comments/requests:

1. we are changing to board/$vendor/$model and soc/$model layout (it's not completed yet because we have to migrate imx6/usbarmory on all linked projects, but we juts added support for mx6ullevk in this fashion). Could you please move paths accordingly (e.g. board/raspberry/pizero, board/raspberry/pi2 and soc/bcm2835) ?

2. Some style/consistency notes with the rest of our code base, these are not defects or mistakes on your part, but we just like the codebase to be consistent:

*  bcm2835.HardwareInit() should be named Init() to be consistent with imx6.Init()

* please note that 

* The `arm` package exports `TimerMultiplier` and `TimerFn` intentionally. We either change this to allow rationale for `InitSpecificTimer` or, my preference, I think this function can be spared and you can just set ARM.TimerMultiplier and ARM.TimerFn when allocating &arm.CPU{} or in bcm2835.Init().

* `PeripheralBase` should not be exported by the package as it is accepted by the initialization function, the initialization function should validate it against common/known ranges if possible (like in this [example](https://github.com/f-secure-foundry/tamago/blob/master/imx6/iomux.go#L66)).

* The comment "care must be taken to ensure that no heap allocation is performed (e.g. defer is not possible)" is [no longer relevant](https://github.com/f-secure-foundry/tamago/commit/3721102e5fbb9968066526e8301926c212c14a9c). Please remove it and defer in a manner consistent with the rest of the package.

* Please convert `Line: 0x2f` in `pi/pizero/board.go` with use of the proper constants used to derive this value (if any can be applied to describe its meaning).

* The purpose of `pi/pizero/tls.s` is unclear, could you clarify its need?

There other style changes or improvements that I'd like to suggest, but for a first round I think this is probably plenty :). In the meantime thank you so much for this contribution, I look forward to merge it.

* Given the [properties of Go constants](https://blog.golang.org/constants) we find that `uint32` can be spared from them as it detracts, in my opinion, from "syntactic sugar" and also results in uint32(foo) casting required when performing pointer arithmetics. Given that the type is inferred by function arguments it is our style to avoid this, please see [this commit](https://github.com/f-secure-foundry/tamago/commit/fc015e08ff59f85977b6be55af66d912f6878ba1).

* `bcm2835/rng.go`, `bcm2835/uart.go` should be consistent with `imx6/rngb.go`, `imx6/uart.go`, and therefore use an instance. This improves the way documentation is presented to the user.

* I wonder if a [busyloop](https://github.com/f-secure-foundry/tamago/blob/master/arm/timer.s#L70) can work better than the `dummy` function in `bcm2835/uart.go`. Also any reason for simply not using time.Sleep() there ?

* I am not sure I understand the usefulness of `var GPIO = gpio{}` given that the struct holds no state/lock (which btw for consistency with the rest of the codebase should be `&gpio{}`).

kenbell:
Thanks for the detailed review, I'll work through these.

I didn't really understand the rationale for using instances vs not in the packages, so i somewhat randomly used them vs not.  If i understand the purpose, it's for documentation and/or when there's state to hold?  I added GPIO so that code can do `pi.GPIO.Set()` - I'll rework to be `pi.SetGPIO()` style.

I exposed PeripheralBase because code outside of the package that is common to both boards needs access to the value.  It's a fundamental base address for all hardware register access in the Pi's.   I'll expose it as `bcm2835.PeripheralAddress(offset)` so that code outside of the bcm2835 package can still access it and remain board agnostic.

If I have any questions on the other items I'll get back - but the all seem clear.

abarisani:
> Thanks for the detailed review, I'll work through these.
> 
> I didn't really understand the rationale for using instances vs not in the packages, so i somewhat randomly used them vs not. If i understand the purpose, it's for documentation and/or when there's state to hold? I added GPIO so that code can do `pi.GPIO.Set()` - I'll rework to be `pi.SetGPIO()` style.
> 
> I exposed PeripheralBase because code outside of the package that is common to both boards needs access to the value. It's a fundamental base address for all hardware register access in the Pi's. I'll expose it as `bcm2835.PeripheralAddress(offset)` so that code outside of the bcm2835 package can still access it and remain board agnostic.
> 
> If I have any questions on the other items I'll get back - but the all seem clear.

I do like pi.GPIO instances but only when they represent an individual GPIO instance rather than the overall GPIO functionality of the Pi. But honestly it can go both ways, maybe add a mux to reason it ;) (which is probably not a bad idea to prevent concurrent operation) ? Or wouldn't an instance that holds information on the line and its configuration make more sense to match more what we are doing [here](https://github.com/f-secure-foundry/tamago/blob/master/imx6/gpio.go) ? I think it's nice if you can have an instance of a configured gpio that then can be set to in/out high/low and that's it, see how we implement leds [here](https://github.com/f-secure-foundry/tamago/blob/master/usbarmory/mark-two/led.go).

It's fine to expose PeripheralAddress, my rationale is that it should be set only at initialization time and if anyone needs to access it then it shall be done through a function so that it's read-only post initialization.

Thanks!

kenbell:
@abarisani - I think I've addressed all comments.  I've rebased due to directory structure change, and done one commit per comment.

Looking again, it might be better to solve `pi/pizero/tls.s` in tamago-go.  In essence, because PiZero support is compiled with GOARM<7, the Go runtime expects the symbol `runtime·read_tls_fallback`.  I can think of three approaches:
1 - what we do here, solve this board specific problem in tamago code-base
2 - add the symbol to the tamago-go code-base
3 - change the conditional logic that requires this symbol in tamago-go code-base (https://github.com/golang/go/blob/master/src/cmd/internal/obj/arm/obj5.go#L69)

I think this symbol will be needed on any SOC where GOARM<=6, so a generic solution may be best.

abarisani:
Can you change the PR so that it targets our development branch rather than master, so that we can keep working on development while we refine these changes?

Overall everything looks ok, again I have some further change proposals but we I can do a PR myself for you to review against the development branch.

Concerning `runtime·read_tls_fallback` I agree this should be handled outside tamago, but rather in tamago-go. To this end you can remove your code related to it as I've just [committed](https://github.com/f-secure-foundry/tamago-go/commit/ab6a561d3d0642ca7f8928c661613f58144dd7f6) its definition in tamago-go1.15.1 (which will be carried on in future releases and branches).

kenbell:
I've moved to dev branch and removed read_tls_fallback symbol.

abarisani:
Merged. Please expect my PR sometime early next week. In the meantime thanks!

abarisani:
@kenbell can you please review https://github.com/f-secure-foundry/tamago/pull/6 ?

PR #7, opened 2020-09-22
typo: whitespace cleanup

prusnak:
This PR removes the unnecessary whitespace.

It can be seen here that only the whitespace changes are introduced: https://github.com/f-secure-foundry/tamago/pull/7/files?&w=1

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 TamaGo - bare metal Go for ARM SoCs
 ===================================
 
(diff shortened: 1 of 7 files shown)


abarisani:
These whitespaces are required in Markdown viewers to insert a newline. Quoting markdown syntax "To add a single extra line after a paragraph, add two extra spaces".

abarisani:
I however addressed the one in the README which is legit, thx.

PR #13, opened 2020-09-25
add Raspberry Pi 1 support

prusnak:
This PR adds Raspberry Pi 1 (B+) support to Tamago.

It's an amalgamation of Pi Zero and Pi 2 code.

Tested on Raspberry Pi 1 B+ v1.2 via https://github.com/prusnak/tamago-example-pi1, which is a slightly modified version of https://github.com/kenbell//tamago-example-pi2 (the only diff is the latest commit)


kenbell:
Nice

The ramSize is getting a bit messy.  For Pi, I think there's a way we can interrogate the VideoCore for the actual available RAM.  As a separate change, I'll take a look if there's a way to hook that in sufficiently early in the startup sequence.

prusnak:
I did some more testing with Raspberry Pi 1 A (not A+) and Pi 1 B (not B+) both from 2011 and here are the results:
* Raspberry Pi 1 B - everything works except for the LEDs, the PCB layout has changes so I assume they are wired to different pins
* Raspberry Pi 1 A - the same issue with the LEDs + the RAM needs to decreased to 256 MB

Raspberry Pi 1 B+ I used for the initial testing works flawlessly.

prusnak:
Seen the videocore stuff in https://github.com/f-secure-foundry/tamago/pull/14, very cool! Here's the output of the functions for boards I have lying around. It seems this "interrogation" can indeed be used to detect the available RAM. However, other values, such a BoardModel/MACAddress/Serial/CPUAvailableDMAChannels look fishy.

Raspberry Pi 1 A

```
FirmwareRevision: 8 (dec)
BoardModel: 0
MACAddress: 0xb827ebc95629
Serial: 0
CPUMemory: 0x0, 0xc000000
GPUMemory: 0xc000000, 0x4000000
CPUAvailableDMAChannels: 0x7f35
```

Raspberry Pi 1 B

```
FirmwareRevision: 13 (dec)
BoardModel: 0
MACAddress: 0xb827ebf4f296
Serial: 0
CPUMemory: 0x0, 0x1c000000
GPUMemory: 0x1c000000, 0x4000000
CPUAvailableDMAChannels: 0x7f35
```

Raspberry Pi 1 B+

```
FirmwareRevision: 0x10
BoardModel: 0
MACAddress: 0xb827eb32d0cc
Serial: 0
CPUMemory: 0x0, 0x1c000000
GPUMemory: 0x1c000000, 0x4000000
CPUAvailableDMAChannels: 0x7f35
```

abarisani:
An alternative to introduce a lot of new code to query available RAM could also be to have different mem.go depending on a build tag which specifies the actual model. This might be cleaner, simpler and introduce less code.

abarisani:
If all LEDs are always going to be treated identically by each model, wouldn't it be better to move the functions repeated in `pizero`, `pi2` and `pi1` to the `pi` package and have each model-specific `Init` just initialize the list of available names, numbers ? This would also simplify the need for the `Board` type which at the moment only supports the LED functions. This came up while reviewing this as there is yet another repetition for those functions. Thoughts @kenbell ?

abarisani:
Any thoughts on my comment?

kenbell:
@abarisani - With the Pi, the user can change the ratio of RAM for the ARM CPUs vs the GPU using the config.txt file read by the firmware: https://www.raspberrypi.org/documentation/configuration/config-txt/memory.md

I don't know if supporting that entirely makes sense in tamago use-cases, but it's something that could conflict with setting fixed values for the ARM RAM available.

abarisani:
I'd like to merge this but I'd need it to target development and not master so that we can stage it in that branch first, thanks!

prusnak:
Rebased on top of `development` branch

abarisani:
This is now also merged in `master`. Thanks for your contribution!

PR #14, opened 2020-09-26
Initial Pi VideoCore and DMA support

kenbell:
* DMA is just memory to memory transfers for now
* VideoCore mailbox is functional


prusnak:
Very cool! Here's the output of the functions for boards I have lying around. It seems this "interrogation" can indeed be used to detect the available RAM like suggested in https://github.com/f-secure-foundry/tamago/pull/13#issuecomment-699247428. 

However, other values, such a BoardModel/MACAddress/Serial/CPUAvailableDMAChannels look fishy.

Raspberry Pi 1 A

```
FirmwareRevision: 8 (dec)
BoardModel: 0
MACAddress: 0xb827ebc95629
Serial: 0
CPUMemory: 0x0, 0xc000000
GPUMemory: 0xc000000, 0x4000000
CPUAvailableDMAChannels: 0x7f35
```

Raspberry Pi 1 B

```
FirmwareRevision: 13 (dec)
BoardModel: 0
MACAddress: 0xb827ebf4f296
Serial: 0
CPUMemory: 0x0, 0x1c000000
GPUMemory: 0x1c000000, 0x4000000
CPUAvailableDMAChannels: 0x7f35
```

Raspberry Pi 1 B+

```
FirmwareRevision: 0x10
BoardModel: 0
MACAddress: 0xb827eb32d0cc
Serial: 0
CPUMemory: 0x0, 0x1c000000
GPUMemory: 0x1c000000, 0x4000000
CPUAvailableDMAChannels: 0x7f35
```

abarisani (changes requested):
* Please rename dmaController to DMAController (it is exported through the `DMA` variable anyway so it's beneficial to export the type as well.

* Rather than having `lock` as sync.Mutex, requiring c.lock.Lock(), please just declare `sync.Mutex` as a struct member so that you can do c.Lock() (see [here](https://github.com/f-secure-foundry/tamago/blob/master/dma/dma.go#L37)). Same thing applies to `mailbox` type.

* To be consistent with the rest of the code `regBase` should be `base` ?

* If `CopyRAMToRAM` is not RAM specific (e.g. would it work on other kind of memories?) it's probably better to rename it to something simpler? Usually in the DMA world MemToMem is used, but I'd also favor the simplicty of `Copy` ?

* The `136` in function `EDID` should be a constant.

* Please maintain the copyright notices format which I added in the last merge. Also the header of `framebuffer.go`, `messages.go` is inconsistent with the others.

* Function `PhysicalSize` has a typo in its comment as well as a spurious newline, it would also be best to name the return arguments to indicate what they are consistently with its set counterpart.

* In `messages.go` why not attaching the const comments to each individual group? Or have the URL as group comment and the individual values as a comment before each constant?

* In `videocore.go` all functions follow the same pattern for msg creation, mailox call, etc. I feel maybe this should be abstracted with a function that takes ID, buffer size and tag as arguments? There is clearly a repeating pattern and the whole file could be much shorter with just one helper function don't you think?

Thanks for the contribution! Looking forward to merge it.

abarisani:
Heya, any thoughts on my change request?

kenbell:
Hey - the comments look good.  Got distracted with another project.  I'll work on making the changes.

kenbell:
I've pushed a change that I think addresses the review comments.

PR #25, opened 2021-05-04
imx6: Fix IOMUXC_END value

connojd:
The IOMUXC range ends at 0x020e3fff. Fix this from 0x0203ffff
and change the range check in NewPad to use && instead of ||
to detect invalid addresses.

### DIFF
--- a/soc/imx6/iomux.go
+++ b/soc/imx6/iomux.go
@@ -18,7 +18,7 @@ import (
 // IOMUX registers
 const (
 	IOMUXC_START = 0x020e0000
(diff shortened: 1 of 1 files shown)


connojd:
Found this issue while porting tamago to the SABRE lite board

abarisani:
Thanks for spotting this!

abarisani:
FYU: we sometimes set daisy to 0 when we do not use it, because of the correct check I see that we were handling this incorrectly and I resolved it in https://github.com/f-secure-foundry/tamago/commit/a2e021e0e2f60e39279c50696cf01936beef3225

PR #41, opened 2023-05-03
Add a bit realistic parameters to CCID descriptor

ckahlo:
Certain CCID drivers (esp. on Windows) will check the descriptor for valid parameters (clock, modes, protocol handling, features, ...) and deny access if the calculated results are out of spec (ISO7816-3).


abarisani:
These feel too specific for TamaGo, I will move them to our main application which uses these (GoKey) to promote application set values.

PR #66, opened 2026-03-18
Add ARMv5 and NUC980 support

MDr164:
This adds the Nuvoton NUC980 and one of its reference boards the NuMaker-IIoT-NUC980 which is a low cost ARM9 (ARMv5) based SoC used in some iBMC aka industrial or 'edge' BMC solutions by several manufacturers. The datasheet and software (u-boot + Linux) are publicly available from Nuvoton. I'll add a guide on how to test this on hardware including logs a bit later once everything works as intended. Opening a draft PR for initial discussion.


arm/exception.go
```diff
@@ -149,15 +157,19 @@ func (cpu *CPU) initVectorTable(vbar uint32) {
...
 
 	// set vector base address register
-	set_vbar(cpu.vbar)
-
-	if cpu.Secure() {
-		// set monitor vector base address register
-		set_mvbar(cpu.vbar)
+	// ARM926EJ-S (ARMv5TEJ) does not have VBAR; vectors are fixed at
+	// 0x00000000. Skip the register write on such cores via NoVBAR.
+	if !cpu.NoVBAR {
```
abarisani: Can't you just check `vbar != 0` rather than having a new boolean?

arm/exception_v5.s
```diff
@@ -0,0 +1,170 @@
...
+									\
+	/* remove exception specific LR offset */			\
+	SUB	$LROFFSET, R14, R14					\
+									\
+	/* save caller registers */					\
+	MOVM.DB.W	[R0-RN, R14], (R13)	/* push {r0-rN, r14} */	\
+									\
+	/* restore g in case this mode banks them */			\
+	MOVW	$SAVE_SIZE, R0						\
+	CMP	$44, R0							\
```
abarisani, line 98: What does 44 represent here?

arm/init_v5.go
abarisani, line 1: I'd rename `arm/init_v5.go` for consistency, in general it would be ideal for all v5 specific code to go under `_v5` prefixes in this package.

arm/irq.go
```diff
@@ -26,29 +26,49 @@ func wfi()
 // EnableInterrupts unmasks IRQ interrupts in the current or saved program
 // status.
 func (cpu *CPU) EnableInterrupts(saved bool) {
+	if cpu.IRQEnableOverride != nil {
```
abarisani: If these overrides are always needed for v5 then we should probably have `irq_v5.go` and avoid these ifs all around.

soc/nuvoton/nuc980/arm926ej.s
abarisani, line 1: Shouldn't this live under arm for v5 support or is it specific to the NUC980?

soc/nuvoton/nuc980/nuc980.go
```diff
@@ -0,0 +1,104 @@
...
+
+	HCLKEN_CRPT = uint32(1 << 23) // CRPT AHB clock enable (datasheet p. 153)
+	PCLKEN0_TMR = uint32(1 << 8)  // Timer0 APB clock enable
+	PCLKEN0_UA0 = uint32(1 << 16) // UART0 APB clock enable
+)
+
+// ARM processor instance
+var ARM = &arm.CPU{
+	TimerMultiplier:    1,
+	NoVBAR:             true,
```
abarisani: See earlier comments, it feels odd for these to be defined here. It should be under `arm` if not specific to the SoC, maybe with a separate `arm.CPUv5` type if that helps?

soc/nuvoton/nuc980/rng.go
abarisani, line 1: Consistently with the rest of the tamago code base this should be a driver under `nuvoton/prng` which exposes a `prng.PRNG` instance (which would also avoid the static variables set here).

Examples [here](https://github.com/usbarmory/tamago/tree/development/soc/microchip/trng) and [here](https://github.com/usbarmory/tamago/tree/development/soc/nxp/rngb).

soc/nuvoton/nuc980/timer.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"github.com/usbarmory/tamago/internal/reg"
+)
+
+// ETimer0 register addresses
+//
+// NUC980 Series Datasheet, p. 185 (§ 6.9 Timer Controller).
+const (
+	ETMR0_BA = 0xB0050000
```
abarisani: Here and everywhere else please use lowercase letters in hex addresses to keep it consistent with the rest of the tamago codebase.

soc/nuvoton/nuc980/timer.go
abarisani, line 1: Consider having this as a driver timer, similar to what has been done for [CLINT](https://github.com/usbarmory/tamago/tree/development/soc/sifive/clint).

As this allow cleaner integration ([example](https://github.com/usbarmory/tamago/blob/development/soc/sifive/fu540/timer.go) possibly and avoid static variables).

soc/nuvoton/nuc980/uart.go
abarisani, line 1: This should live in `nuvoton/uart` consistently with other SoCs in tamago.

abarisani:
Haven't forgotten about this, just a little swamped but I promise I'll get to it.

MDr164:
No worries, I just noticed a merge conflict with the development branch so I just updated the code to match the past commits. I can at least report that the code has been working well so far and I'm right now working on more drivers and the other PR so I'm also busy in the meantime :)

abarisani:
This week I have time to work on this, can you see and address the pending comments? There have been also some changes that need sync up. Thanks!

soc/nuvoton/nuc980/aic.go
abarisani, line 1: If this is ever re-used in Nuvoton SoCs it should live under `soc/nuvoton/aic` consistently with the rest of the codebase.

soc/nuvoton/nuc980/aic.go
```diff
@@ -0,0 +1,110 @@
...
+	reg.Write(REG_AIC_MDCRH, 0xFFFFFFFF)
+}
+
+// EnableIRQ unmasks interrupt source irq (0..63).
+func EnableIRQ(irq int) {
+	if irq < 0 || irq > 63 {
+		return
+	}
+	if irq < 32 {
+		reg.Write(REG_AIC_MECR, 1<<uint(irq))
```
abarisani: Write takes an uint32 as argument, I haven't tested it but I wonder if this compiles at all?

MDr164:
Quick summary once you get around reviewing it: I currently use a patched tamago-go toolchain based on your 1.26.2 tag with one commit on top: https://github.com/MDr164/tamago-go/commit/f4bcfd42e253b4c4d64bc2e4bd0344ac9cf73c3d

Additionally I created a small script that I run locally to create correctly formatted sd-card images for this platform. I have not tested booting it from its internal NAND flash yet but that should work with the current code just fine I assume. I did not push the script into this PR as it felt to be out of scope to me.

One thing that tripped me up about GOARM buildtags is that they are apparently cumulative which the doc.go file they are described in doesn't claim. So:
GOARM=5 ~= arm.5
GOARM=6 ~= arm.5 || arm.6
GOARM=7 ~= arm.5 || arm.6 || arm.7

That's why I opted for using !arm.6 instead of going for the arm.5 buildtag.

soc/nuvoton/nuc980/aic.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+// DisableAll masks all 64 AIC interrupt sources.
+func DisableAll() {
+	reg.Write(REG_AIC_MDCR, 0xFFFFFFFF)
+	reg.Write(REG_AIC_MDCRH, 0xFFFFFFFF)
+}
+
+// EnableIRQ unmasks interrupt source irq (0..63).
+func EnableIRQ(irq int) {
```
abarisani: Consistently with the rest of the tamago code base please use `EnableIRQ`

soc/nuvoton/nuc980/aic.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	if irq < 32 {
+		reg.Write(REG_AIC_MECR, 1<<uint(irq))
+	} else {
+		reg.Write(REG_AIC_MECRH, 1<<uint(irq-32))
+	}
+}
+
+// DisableIRQ masks interrupt source irq (0..63).
+func DisableIRQ(irq int) {
```
abarisani: Consistently with the rest of the tamago code base please use `DisableIRQ`

soc/nuvoton/nuc980/earlyinit.s
```diff
@@ -0,0 +1,40 @@
...
+
+// EarlyInit configures NUC980 clock gates from assembly.  Writes to
+// the SYS/CLK register block (0xB0000000-0xB00003FF) silently fail
+// when performed via the Go runtime's store path; direct STR from
+// early assembly works reliably.
+
+#include "textflag.h"
+
+// func EarlyInit()
+TEXT ·EarlyInit(SB),NOSPLIT|NOFRAME,$0
```
abarisani: This deserves a better name, ClockInit ?

soc/nuvoton/nuc980/rng.go
MDr164: I wasn't sure if other Nuvoton SoCs actually use the same IP or if it's specific to the NUC9x0 type of chips. But sure, I can move that code out to a more common place (same for other IP drivers which I'm not sure if they are Nuvoton or SoC specific as this is the only Nuvoton chip I got here).

soc/nuvoton/nuc980/nuc980.go
MDr164: It's as far as I know not specific to the ARMv5 core class (I actually got some more ARM targets here one being v5 the other v6 and both lack the VBAR register). But I assume we can do that without the extra bool and simply check `cpu.vbar != 0` in the other code section. I'll try to refactor it and test on hardware again.

soc/nuvoton/nuc980/arm926ej.s
MDr164: It is not specific to the NUC980 indeed, I'll move it out 👍

soc/nuvoton/nuc980/aic.go
MDr164: It does compile and run just fine on my board. I assume it's just cast into uint32

soc/nuvoton/nuc980/earlyinit.s
MDr164: It doesn't enable all clocks, just the ones that are required to be up very early. Maybe something like EarlyClockInit then? I do kinda wanna keep the "early" part in there to note that this is required before a lot of other tasks can run.

MDr164:
I'll update the code by moving everything to a common place for drivers inside the nuvoton package. I assume if there are other nuvoton chips using different IP cores we can still move things around. Do you think the one commit I got for the Go toolchain is worth adding to tamago-go? I guess if I send it upstream there would be confusion about why I run into this edgecase as it only really affects baremetal targets which aren't upstream yet.

soc/nuvoton/nuc980/aic.go
abarisani: The shift operation does the trick, I wasn't aware of this. It would have not work without it.

abarisani:
Yes, for starters we can add that commit to tamago-go, I'll ponder it a little more but in time for v1.26.3.

abarisani:
Sorry for the delay, I am testing your tamago-go patch for inclusion, however it fails to pass this test (it hangs):

```
GOOS=tamago GOARCH=arm GOARM=5 bin/go test sync/atomic internal/sync internal/runtime/atomic
```

abarisani:
I think the issue spawns from the fact that the testsuite is not designed for single-core processors without atomic operations, I had to workaround this issue [here](https://github.com/abarisani/tamago-go/blob/9dafdf81964526df126717ec95d9aa1bbfa9769c/src/runtime/debug.go#L78) to enable lack of atomics on smaller RISC-V targets for [kotama](https://github.com/usbarmory/kotama).

I think a similar workaround is needed though I am slightly uncomfortable in having the same approach for upstreaming it as I'd favor a conditional patch which is never touched on other architectures.

MDr164:
> Sorry for the delay, I am testing your tamago-go patch for inclusion, however it fails to pass this test (it hangs):
> 
> ```
> GOOS=tamago GOARCH=arm GOARM=5 bin/go test sync/atomic internal/sync internal/runtime/atomic
> ```

I can take a look and rebase the PR as well against the dev branch again, I also meant to get back sooner to it but this soc changed places with another one here internally which I planned to upstream later but had to be ready until August as I'm giving a talk about Tamago then and we changed demo targets. Now that the demo is ready I can get back to this one here.

MDr164:
> I think a similar workaround is needed though I am slightly uncomfortable in having the same approach for upstreaming it as I'd favor a conditional patch which is never touched on other architectures.

Fair, I can take a look if I come up with a solution quicker than you (which I doubt). I got a second target here on the roadmap that uses the exact same cpu core as this soc so it'll at least not be a one-time thing for arm

MDr164:
fyi: rebased and tested on hardware again:
```sh
Runtime   : go1.26.4 tamago/arm
Board     : NuMaker-IIoT-NUC980G2
SoC       : NUC980 PDID 0x0
CPU       : ARM926EJ-S MIDR 0x41069265
  impl    : 0x41 (ARM)
  variant : 0x0
  arch    : 0x6
  part    : 0x926
  rev     : 0x5
RAM       : 128 MB @ 0x00000000
ETimer1   : CTL=0x0 DR=0 ISR=0x0
ETimer1   : IER=0x1 CMPR=9999
AIC       : ISNR=0
Starting ServiceInterrupts...
Waiting 50ms for relay goroutine to initialize...
ETimer1 started.
Idle governor: WFI

tick 0 (ETMR1_DR=2130 ISR=0x0)
tick 1 (ETMR1_DR=1777 ISR=0x1)
tick 2 (ETMR1_DR=1777 ISR=0x1)
tick 3 (ETMR1_DR=1777 ISR=0x1)
tick 4 (ETMR1_DR=1776 ISR=0x1)
tick 5 (ETMR1_DR=1777 ISR=0x1)
tick 6 (ETMR1_DR=1777 ISR=0x1)
tick 7 (ETMR1_DR=1882 ISR=0x1)
tick 8 (ETMR1_DR=1776 ISR=0x1)
tick 9 (ETMR1_DR=1777 ISR=0x1)
tick 10 (ETMR1_DR=1784 ISR=0x1)
```

I think I covered all the review comments now, will go over it again the next couple days but I'm pretty certain it works well now. I'll try to arrange getting hardware to you for this target as well.

(side note: I also update the armcas toolchain patch to 1.26.4 as that one is still needed)

[0001-NOT-FOR-UPSTREAM-NUC980-example.patch](https://github.com/user-attachments/files/28640173/0001-NOT-FOR-UPSTREAM-NUC980-example.patch) is used for testing on the board

abarisani:
I think I resolved all addressed comments and left open the remaining ones (though I could be wrong as I need to re-check the patch again), please give me a hint on when I can do a final review. Thanks!

arm/exception_v5.s
MDr164: Added a comment above that section, not entirely sure if that is the best way to achieve detection but it at least works reliably well

soc/nuvoton/nuc980/nuc980.go
MDr164: I removed NoVBAR entirely now and only added the check in some very specific cases, namely arm.go and exception.go directly

soc/nuvoton/nuc980/rng.go
MDr164: I moved the driver code, I'm still not really sure how reusable it's going to be in the end but no harm moving it and aligning with the rest

arm/init_v5.go
MDr164: Done

soc/nuvoton/nuc980/arm926ej.s
MDr164: Done

MDr164:
> Please give me a hint on when I can do a final review. Thanks!

I did some minor cleanups of some code comments now and double checked the code. It's in a state now where I'd call it final.

abarisani:
I am reviewing modified files and new files.

I completed the review of modified files, it looks fine but I felt the need to keep consistency with the current code base in terms of flow, constant re-use, comments.

Please see the attached pass to address these, I will add an eventual patch on new files later today.

Thanks!

[nuc980.patch](https://github.com/user-attachments/files/28701335/nuc980.patch)

MDr164:
Thanks a lot for the review so far! I'll take a look at your patch and hold out on updating this branch until you went over the other files to not create too much noise for now.

I also had a look at the test failure for the single core no-atomic ARMv5 addition I had to make. I suppose as the only current expected outcome upstream in Go is a helper from e.g. Linux capturing that case we might be able to simply skip that particular case in that particular configuration similarly to how the GOSOFT flag you added had to make an exception (as you mentioned). Not ideal but I also don't see any better option right now.

abarisani:
If you are happy with the patch I can push it to your branch directly. Let me know.

MDr164:
It's dropping a couple of comments here and there which I would have kept but that is very subjective as I prefer a bit of context when reading assembly as I can't memorize every detail especially when jumping between arm, arm64 and riscv(32/64) daily. As it doesn't alter the codes functionality I'm still fine with shrinking the amount of comments though. :+1:

board/nuvoton/nuc980iiot/console.go
```diff
@@ -0,0 +1,27 @@
...
+	_ "unsafe"
+
+	"github.com/usbarmory/tamago/soc/nuvoton/nuc980"
+)
+
+//go:linkname printk runtime/goos.Printk
+func printk(c byte) {
+	// Emit CR before LF so output is legible on serial terminals that do
+	// not perform implicit CR insertion (raw mode, minicom, screen, etc.).
+	if c == '\n' {
```
MDr164: Actually re-reading this, it might not be the best place to insert them here but rather one layer higher up. This here was just a result of changing devtools and suddenly seeing odd prints.

abarisani:
> It's dropping a couple of comments here and there which I would have kept but that is very subjective as I prefer a bit of context when reading assembly as I can't memorize every detail especially when jumping between arm, arm64 and riscv(32/64) daily. As it doesn't alter the codes functionality I'm still fine with shrinking the amount of comments though. :+1: 

Any specific comment you would like to keep.

I considered the ones I removed self explanatory tbh but happy to restore anything you think it's necessary.

By the way I assumed you did yet the changes against your target?

MDr164:
> Any specific comment you would like to keep.

The one above EXCEPTION_V5 for example, I could see that we could shrink it but fully dropping it would require reading the asm again.

> By the way I assumed you did yet the changes against your target?

Yes and it currently doesn't build with your patch due to moving ServiceInterrupts to *CPU. I added the TODO comment as it wouldn't be needed in this PR; and it affects every arm target afaik :sweat_smile: 

Let me quickly update my code to use ARM.CPU.ServiceInterrupts to test after your patch but I wouldn't see how updating comments and moving the if conditions around for vbar affects the board.

MDr164:
I recompiled and caught one small issue in your patch: in irq_v5.s line 77 you used a register that doesn't exist on ARMv5. I changed C12 back to C7 and everything was back to normal.

abarisani:
Thanks. Yes the CPU move is an API change, it's intentional.

MDr164:
I applied your patch, fixed the wrong register in it and added one comment you deleted back in a shorter form now.

abarisani:
> I recompiled and caught one small issue in your patch: in irq_v5.s line 77 you used a register that doesn't exist on ARMv5. I changed C12 back to C7 and everything was back to normal.

Yep, typo on my side. Sorry for that.

abarisani:
I took the liberty for a small clean up at https://github.com/MDr164/tamago/commit/70bc6bc1d709012c0a2f6eddda2e79f4adf18d31 (feel free to revert if inappropriate).

soc/nuvoton/etimer/etimer.go
```diff
@@ -0,0 +1,110 @@
...
+// SetCompare sets the 24-bit compare value at which the timer wraps or
+// triggers its interrupt.
+func (hw *ETimer) SetCompare(compare uint32) {
+	reg.Write(hw.Base+CMPR, compare&CMPR_MAX)
+}
+
+// EnableInterrupt enables or disables the compare-match interrupt.
+func (hw *ETimer) EnableInterrupt(enable bool) {
+	if enable {
+		reg.Write(hw.Base+IER, 0x1)
```
abarisani: Throughout tamago we use the `bits` or `reg` package to set or clear the relevant bits, which are always added as a constant, with Set/SetTo as needed.

This thereforce could be:

```
reg.SetTo(hw.Base+IER, IER_LVD_IEN, enable)
```

Please consider being more consistent to the rest of tamago drivers, which avoid using Read/Write for partial read/writes.

soc/nuvoton/nuc980/earlyinit.s
```diff
@@ -0,0 +1,54 @@
...
+// the SYS/CLK register block (0xb0000000-0xb00003ff) silently fail
+// when performed via the Go runtime's store path; direct STR from
+// early assembly works reliably.
+
+#include "textflag.h"
+
+// func EarlyClockInit()
+TEXT ·EarlyClockInit(SB),NOSPLIT|NOFRAME,$0
+	// CLK_HCLKEN (0xb0000210): enable AIC AHB clock (bit 10).
+	MOVW	$0xb0000210, R0
```
abarisani: Here and later these should be constants, either in .s or .go files depending on whether there is desire to export them or not.

soc/nuvoton/nuc980/nuc980.go
```diff
@@ -0,0 +1,149 @@
...
+const (
+	// SYS_BA is the System Manager / Global Control register base.
+	SYS_BA = 0xb0000000
+
+	// SYS_PDID is the Part Device Identification register (read-only).
+	// Returns the chip part number, e.g. 0x00098000 for NUC980.
+	SYS_PDID = SYS_BA + 0x000
+
+	// SYS_REGWPCTL is the register write-protection control register.
+	// Writing the three-byte unlock sequence {0x59, 0x16, 0x88} in order
```
abarisani: This comment provides an implementation insight which I feel is unnecessary to godoc readers.

soc/nuvoton/nuc980/nuc980.go
```diff
@@ -0,0 +1,149 @@
...
+// MIDR returns the ARM Main ID Register.
+func MIDR() uint32 {
+	return ARM.MIDR()
+}
+
+// SoftReset triggers an immediate chip reset via the AHB IP reset register.
+// The function does not return; the CPU restarts from the boot ROM.
+func SoftReset() {
+	// Unlock SYS register write protection.
+	for reg.Read(SYS_REGWPCTL)&1 == 0 {
```
abarisani: reg.Get would be more idiomatic here.

soc/nuvoton/nuc980/nuc980.go
```diff
@@ -0,0 +1,149 @@
...
+		reg.Write(SYS_REGWPCTL, 0x59)
+		reg.Write(SYS_REGWPCTL, 0x16)
+		reg.Write(SYS_REGWPCTL, 0x88)
+	}
+	// Trigger chip reset (bit 0 = CHIPRST).
+	reg.Write(SYS_AHBIPRST, 0x1)
```
abarisani: reg.Set would be more idiomatic here

soc/nuvoton/nuc980/nuc980.go
```diff
@@ -0,0 +1,149 @@
...
+	for {
```
abarisani: I think `select {}` might be nicer to CPU usage here (which I understand is quite short, but still from a style perspective it feels nicer).

MDr164:
Thanks, the changes are reasonable. I'm currently rethinking if I can move some code from cpuinit.s which is in the board package to the earlyinit.s of the soc package instead. The actual content would not change, just the location.

soc/nuvoton/nuc980/earlyinit.s
```diff
@@ -0,0 +1,54 @@
...
+	MOVW	(R0), R1
+	ORR	$(1<<10), R1
```
abarisani: Same applies to values as some of them are already defined as constants but not used? (such as `PCLKEN0_TMR0`)

soc/nuvoton/nuc980/rng.go
```diff
@@ -0,0 +1,42 @@
...
+
+// PRNG is the NUC980 CRYPTO engine pseudo random number generator.
+//
+// TODO: replace the fixed Seed with a genuine entropy source (e.g.
+// uninitialized SRAM noise, ADC least-significant bits, or a boot-time
+// counter latched before DDR init) so that every boot produces a
+// different PRNG sequence.
+var PRNG = &prng.PRNG{
+	Base: CRPT_BA,
+	Seed: 0xdeadbeef,
```
abarisani: Use a high resolution timer and/or the chip unique identifier here?

soc/nuvoton/nuc980/timer.go
```diff
@@ -0,0 +1,112 @@
...
+	ETimer0 = &etimer.ETimer{Base: ETMR0_BA}
+	ETimer1 = &etimer.ETimer{Base: ETMR1_BA}
+)
+
+// timerWrapPeriod is the number of ticks in one full counter cycle.
+const timerWrapPeriod = uint64(etimer.CMPR_MAX) + 1
+
+// timerLast holds the most recent hardware counter value seen by
+// readTimerExtended.
+var timerLast uint32
```
abarisani: Wrap timerLast/timerHigh in a single `var ( .. )` block?

soc/nuvoton/prng/prng.go
```diff
@@ -0,0 +1,88 @@
...
+
+// GetRandomData fills b with bytes from the CRYPTO PRNG. Each generation
+// produces 256 bits (8×32-bit); the generation is repeated as needed for
+// requests larger than 32 bytes.
+func (hw *PRNG) GetRandomData(b []byte) {
+	read := 0
+	need := len(b)
+
+	for read < need {
+		ctl := uint32(CTL_KEYSZ256) | CTL_START
```
abarisani: Here and below using `bits` package on ctl would be more idiomatic.

soc/nuvoton/uart/uart.go
```diff
@@ -0,0 +1,78 @@
...
+// Init configures the UART for 8N1 operation at the divisor set in Baud.
+//
+// Pin mux and the UART APB clock must be configured by the caller (e.g. in
+// the board cpuinit) before Init.
+func (hw *UART) Init() {
+	if hw.Base == 0 {
+		panic("invalid UART instance")
+	}
+
+	reg.Write(hw.Base+LCR, lcr8N1)
```
abarisani: Using bits/reg for setting bits would be more idiomatic.

abarisani (commented):
Things looks good, I added some style comments for the sake of consistency with the rest of tamago (I understand you might agree or not with some of them). Thanks!

abarisani:
I took the liberty of adding missing documentation in https://github.com/MDr164/tamago/commit/a926b3c19c794fe6f8ffe742cf616de3247ec593

MDr164:
Thanks for the feedback, I applied your suggestions and also reorganized the asm as mentioned by moving everything out of the board pkg except the pinmux which is specific to this board. For some reason doing the pinmux and the EarlyClockInit not via asm it fails to apply correctly. The new code has also been validated on my board here.

abarisani:
`pinmux.s` still could use constants for register address and value, I'd still mux nanotime with the chip UID to have better seed value. The rest looks good and I'll be happy to merge it!

MDr164:
I can change pinmux.s to use constants, sure. For the seed value of the PRNG peripheral of the soc I don't think it is worth using the UID. It is not really unique in that sense as all NUC980 chips share the same value from what I read in the datasheet. The closely related NUC970 would have the same peripheral but a different ID for that matter but I'm not sure if having a chip sku specific seed provide any benefit over simply using the timer value. I think even better would be to query one of the ADCs but I think those are also connected on the pinmux only and not internally to some temp sensor of the chip itself.

abarisani:
> I can change pinmux.s to use constants, sure. For the seed value of the PRNG peripheral of the soc I don't think it is worth using the UID. It is not really unique in that sense as all NUC980 chips share the same value from what I read in the datasheet. The closely related NUC970 would have the same peripheral but a different ID for that matter but I'm not sure if having a chip sku specific seed provide any benefit over simply using the timer value. I think even better would be to query one of the ADCs but I think those are also connected on the pinmux only and not internally to some temp sensor of the chip itself.

I don't quite understand if the PRNG becomes predictable with the same Seed or not.

If so we might want to consider panicking if Seed is empty and delegate that decision to board packages.

MDr164:
> I think a similar workaround is needed though I am slightly uncomfortable in having the same approach for upstreaming it as I'd favor a conditional patch which is never touched on other architectures.

I looked deeper into this and found an instruction that is available on real hardware as well as on Qemu that doesn't cause issues. This commit here now works on the actual board and on Qemu. I added the rational in the code as a lenghty comment which you're not a fan of so feel free to strip it out: https://github.com/MDr164/tamago-go/commit/b32a1322f518d47922b7ead870bbcf657de397c8

MDr164:
> I don't quite understand if the PRNG becomes predictable with the same Seed or not.

The datasheet is not really very descriptive to be honest so maybe it is.

> If so we might want to consider panicking if Seed is empty and delegate that decision to board packages.

I think that would be a fair compromise I'm willing to make, just have the user provide something (e.g. an ADC readout if they know their board as the pinux set with ADC inputs).

abarisani:
> > I think a similar workaround is needed though I am slightly uncomfortable in having the same approach for upstreaming it as I'd favor a conditional patch which is never touched on other architectures.
> 
> I looked deeper into this and found an instruction that is available on real hardware as well as on Qemu that doesn't cause issues. This commit here now works on the actual board and on Qemu. I added the rational in the code as a lenghty comment which you're not a fan of so feel free to strip it out: [MDr164/tamago-go@b32a132](https://github.com/MDr164/tamago-go/commit/b32a1322f518d47922b7ead870bbcf657de397c8)

Please use Go assembly for the `ORR, R0, 0xc0` instruction rather than `WORD`. I see that Go stdlib has `SWPW R3, (R7), R9` so kindly check if that's available as well.

Please make a PR against https://github.com/usbarmory/tamago-go/tree/tamago1.26.4 for testing and inclusion in the next release, also please consider making an upstream PR to Go given that `GOARM=5` is officially supported and I feel this is not specific to `GOOS=tamago` ?

abarisani:
> > I don't quite understand if the PRNG becomes predictable with the same Seed or not.
> 
> The datasheet is not really very descriptive to be honest so maybe it is.
> 
> > If so we might want to consider panicking if Seed is empty and delegate that decision to board packages.
> 
> I think that would be a fair compromise I'm willing to make, just have the user provide something (e.g. an ADC readout if they know their board as the pinux set with ADC inputs).

For now I'd simply use `nanotime` in the SoC package, which is required as you need to start timers first, only if `PRNG.Seed == 0`, so that a previous board override is respected.

abarisani:
Additionally I think PRNG should allow re-seeding at runtime, so probably `PRNG.Seed` should become a non-exported field to have `PRNG.Seed()` to set it and negate `seeded`.

MDr164:
> Also please consider making an upstream PR to Go given that `GOARM=5` is officially supported and I feel this is not specific to `GOOS=tamago`?

Yes, it's an upstream issue but as my commit message mentioned it current doesn't show up upstream as Linux has a helper to mitigate the problem. Tamago/baremetal would be the first user of this path. While I agree that it should land upstream for sure I'm not 100% certain if it's accepted if there are not situations upstream where this assembly fixes an issue other than in baremetal environments. I can try though.

MDr164:
> > > I don't quite understand if the PRNG becomes predictable with the same Seed or not.
> > 
> > 
> > The datasheet is not really very descriptive to be honest so maybe it is.
> > > If so we might want to consider panicking if Seed is empty and delegate that decision to board packages.
> > 
> > 
> > I think that would be a fair compromise I'm willing to make, just have the user provide something (e.g. an ADC readout if they know their board as the pinux set with ADC inputs).
> 
> For now I'd simply use `nanotime` in the SoC package, which is required as you need to start timers first, only if `PRNG.Seed == 0`, so that a previous board override is respected.

Makes sense, the vendors Linux driver simply uses the kernel jiffies as a seed with no way to change/override it. So in some sense we got parity but I agree that we could do better.

abarisani:
> > Also please consider making an upstream PR to Go given that `GOARM=5` is officially supported and I feel this is not specific to `GOOS=tamago`?
> 
> 
> 
> Yes, it's an upstream issue but as my commit message mentioned it current doesn't show up upstream as Linux has a helper to mitigate the problem. Tamago/baremetal would be the first user of this path. While I agree that it should land upstream for sure I'm not 100% certain if it's accepted if there are not situations upstream where this assembly fixes an issue other than in baremetal environments. I can try though.

Then let us keep it as part of our upstreaming proposal given that it's only specific to us. Thanks for clarifying.

I assume there are no !Linux GOARM=5 supported platforms upstream then?

abarisani:
> > > > I don't quite understand if the PRNG becomes predictable with the same Seed or not.
> 
> > > 
> 
> > > 
> 
> > > The datasheet is not really very descriptive to be honest so maybe it is.
> 
> > > > If so we might want to consider panicking if Seed is empty and delegate that decision to board packages.
> 
> > > 
> 
> > > 
> 
> > > I think that would be a fair compromise I'm willing to make, just have the user provide something (e.g. an ADC readout if they know their board as the pinux set with ADC inputs).
> 
> > 
> 
> > For now I'd simply use `nanotime` in the SoC package, which is required as you need to start timers first, only if `PRNG.Seed == 0`, so that a previous board override is respected.
> 
> 
> 
> Makes sense, the vendors Linux driver simply uses the kernel jiffies as a seed with no way to change/override it. So in some sense we got parity but I agree that we could do better.

By the way FIPS mode is implemented under TamaGo, just like in Go, and uses memory jiffies for entropy. Though this is for compliance and I cannot comment on its security, still an option for users to enable.

MDr164:
> Then let us keep it as part of our upstreaming proposal given that it's only specific to us.

Well it's specific to non-linux armv5 which tamago seems to be the first candidate.

> I assume there are no !Linux GOARM=5 supported platforms upstream then?

Looks like it, or other operating systems also install helpers to catch the instructions that wouldn't work standalone but I only saw Linux mentioned upstream in the source.

MDr164:
Added a commit to allow calling `*PRNG.Seed(seed uint32)` at runtime and seeded using nanotime by default now and also the toolchain commit. I think that covers everything now?

soc/nuvoton/prng/prng.go
```diff
@@ -75,7 +74,7 @@ func (hw *PRNG) GetRandomData(b []byte) {
 		bits.Set(&ctl, CTL_START)
 
 		if !hw.seeded {
```
abarisani: Why not moving this code to `Seed` and panic here if hw.seed is 0 ? After all `Seed` should be called by `InitRNG` before its first use (and if not it is a bug), tamago already panics in RNGs that are misconfigured as it treats RNG failures as critical.

soc/nuvoton/nuc980/rng.go
```diff
@@ -33,6 +31,11 @@ func initRNG() {
 	// than in nuc980.Init() to avoid a busy-wait on an unclocked peripheral.
 	reg.Or(REG_CLK_HCLKEN, HCLKEN_CRPT)
 
-	PRNG.Init()
+	// InitRNG also fires before Hwinit1, so start the timer here to make
+	// nanotime() usable as a seed source.
+	initTimer()
+	seed := uint32(nanotime())
```
abarisani: I think this variable is redundant and this can be passed directly as argument to `Seed` ?

soc/nuvoton/prng/prng.go
```diff
@@ -47,19 +47,18 @@ const keyRegs = 8
...
-	if hw.Base == 0 {
-		panic("invalid PRNG instance")
-	}
+// Seed sets the value used to seed the CRYPTO PRNG. The CRYPTO PRNG is
+// deterministic, so its output sequence is fully determined by this value; the
+// seed is loaded on the next GetRandomData call. Seed can be called again at
+// any time to re-seed a running PRNG.
+func (hw *PRNG) Seed(seed uint32) {
+	hw.seed = seed
+	hw.seeded = false
```
abarisani: I think it would be safe to treat a zero seed as invalid and spare the boolean.

soc/nuvoton/prng/prng.go
MDr164: Fair, I'll change it to a panic then

soc/nuvoton/prng/prng.go
MDr164: The reason I added it like this for the GetRandom function to re-seed. If the boolean is set to false it will re-seed with whichever value the seed is currently at. I could also omit the boolean and simply re-seed without any GetRandom calls but I wasn't too sure what will happen if two goroutines compete with either seeding and retrieving data so I coupled the seeding as part of the GetRandom call. That's the only reason for this bool

soc/nuvoton/nuc980/rng.go
MDr164: Done

soc/nuvoton/prng/prng.go
MDr164: Done

soc/nuvoton/prng/prng.go
MDr164: I left the bool and clarified its use be renaming it and making the control flow more obvious.

abarisani: I am not sure I get why `Seed` cannot be entirely responsible for the re-seeding without `GetRandom` being aware of that? Assuming there is always an initial seed, otherwise it would panic, I think the race between Seed and GetRandom is non-existent on a single core.

MDr164: Done, updated to decouple Seed() and GetRandomData() now.

abarisani:
Merged, thanks for your patience in addressing my review and this contribution overall!

MDr164:
> Merged, thanks for your patience in addressing my review and this contribution overall!

Likewise thanks for the reviews and patience with my iterations. With the learnings I'll make sure the next ports will be much more smooth :)

abarisani:
We broke [GoTEE](http://github.com/usbarmory/GoTEE-example) by initializing features after the vector table.

Can you test if this patch doesn't break ARMv5/NUC980 please, it fixes GoTEE and it seems to me the best way to go about this:

```diff
diff --git a/arm/arm.go b/arm/arm.go
index cbae403..a82e4ba 100644
--- a/arm/arm.go
+++ b/arm/arm.go
@@ -86,8 +86,8 @@ func (cpu *CPU) Init() {
 	goos.Exit = exit
 	goos.Idle = cpu.DefaultIdleGovernor
 
-	cpu.initVectorTable()
 	cpu.initFeatures()
+	cpu.initVectorTable()
 }
 
 // Mode returns the processor mode.
diff --git a/arm/features.go b/arm/features.go
index da6781b..491ecb9 100644
--- a/arm/features.go
+++ b/arm/features.go
@@ -22,6 +22,11 @@ const (
 	ID_PFR1_GENERIC_TIMER_MASK     = 0xf0000
 )
 
+// ARM architecture versions
+const (
+	MIDR_ARCH_ARMV6 = 0b0111
+)
+
 // defined in features.s
 func read_midr() uint32
 func read_idpfr0() uint32
@@ -36,7 +41,7 @@ func (cpu *CPU) MIDR() uint32 {
 func (cpu *CPU) initFeatures() {
 	// Do not set read features on cores with fixed exception vectors (e.g.
 	// ARMv5) which have no CP15 feature ID registers.
-	if cpu.vbar == 0 {
+	if read_midr() < MIDR_ARCH_ARMV6 {
 		return
 	}
 
```

MDr164:
The MIDR call returns a value that is higher for some reason on the NUC980. Let me quickly try to create a better comparison check in `initFeatures` asap to get GoTEE fixed again.

EDIT ok found it, the MIDR value contains several fields and we only need to compare bits 16 to 19 and not all of them: https://arm.jonpalmisc.com/latest_sysreg/AArch32-midr

PR #67, opened 2026-03-18
Add FSL91030 and MilkV Vega support

MDr164:
This adds another RISCV64 target namely the MilkV Vega and its SoC. There have been a bunch of resources been uploaded to the milkv-vega GitHub org including the datasheets and reference code namely freeloader (small assembly program at the reset vector, OpenSBI, u-boot and Linux). Some other SoC control like items used to configure the switch fabric are largely done in a closed source userspace binary but all the register definitions are available from the datasheets. This PR remains in draft until everything has been validated to work correctly on hardware and scope of the port has been fully clarified.

NOTE: The original PCB only comes with 4MByte flash, I upgraded mine to 64MByte flash. Tamago does fit into 4MByte but in order to make more use of the platform I changed the flash size.

### DIFF
--- a/board/nuclei/eval_soc/timer_riscv64.s
+++ b/board/nuclei/eval_soc/timer_riscv64.s
@@ -1,0 +1,15 @@
+// Nuclei EvalSoC emulator support for tamago/riscv64
+// https://github.com/usbarmory/tamago
+//
+// Copyright (c) The TamaGo Authors. All Rights Reserved.
+//
+// Use of this source code is governed by the license
+// that can be found in the LICENSE file.
+
+#include "textflag.h"
+
+// func rdtime() uint64
+TEXT ·rdtime(SB),NOSPLIT,$0-8
+	WORD	$0xc0102573	// rdtime a0 (csrr a0, time)
+	MOV	X10, ret+0(FP)
+	RET
--- a/riscv64/irq.s
+++ b/riscv64/irq.s
@@ -21,6 +21,14 @@ TEXT ·irq_enable(SB),NOSPLIT|NOFRAME,$0
 
 	RET
 
+// func meie_enable()
+TEXT ·meie_enable(SB),NOSPLIT|NOFRAME,$0
+	// enable machine level external interrupts (PLIC sources)
+	MOV	$(1<<11), T0		// set MIE.MEIE
+	CSRRS	T0, MIE, ZERO
+
+	RET
+
 // func irq_disable()
 TEXT ·irq_disable(SB),NOSPLIT|NOFRAME,$0
 	// disable machine level software interrupts
--- a/riscv64/riscv64.go
+++ b/riscv64/riscv64.go
@@ -47,7 +47,7 @@ func exit(int32)
 func (cpu *CPU) DefaultIdleGovernor(pollUntil int64) {
 	// we have nothing to do forever
 	if pollUntil == math.MaxInt64 {
-		exit(0)
+		cpu.WaitInterrupt()
 	}
 }
 
--- a/soc/fisilink/fsl91030/boot_riscv64.s
+++ b/soc/fisilink/fsl91030/boot_riscv64.s
@@ -1,0 +1,248 @@
+// Fisilink FSL91030 hardware boot initialization (linkcpuinit)
+// https://github.com/usbarmory/tamago
+//
+// Copyright (c) The TamaGo Authors. All Rights Reserved.
+//
+// Use of this source code is governed by the license
+// that can be found in the LICENSE file.
+//
+// This file overrides the default riscv64 cpuinit when built with the
+// 'linkcpuinit' build tag, performing the hardware initialization (DDR,
+// cache, QSPI clock, FPU, stack) that a first-stage boot loader would
+// otherwise do. It suits loading the image directly into DRAM (e.g. via
+// JTAG); for NOR boot a relocating stage is required instead.
+//
+// The register offsets, the cache-enable value and the DDR controller
+// timing/configuration values are taken verbatim from the Nuclei freeloader.S
+// (Copyright Nuclei System Technologies) shipped with the FSL91030 vendor SDK.
+// The DDR timing values are opaque vendor-provided constants with no documented
+// bit-level meaning and are therefore reproduced as-is; the offsets they are
+// written to are named below for clarity.
+//
+// Sequence: disable cache, initialize DDR, configure the QSPI0 clock,
+// enable cache, disable interrupts and enable the FPU, set the stack to the
+// top of DRAM and jump to the runtime entry. No stack is available during
+// DDR init; register usage:
+//   X14 (A4): DDR controller base, constant
+//   X15 (A5): working register, DDR config values
+//   X12 (A2): 8192, used for multiple timing registers
+//   X13 (A3): derived timing value, reused for REG3
+//   X11 (A1): working register, more timing/config values
+
+//go:build linkcpuinit
+
+#include "textflag.h"
+
+// Nuclei vendor-specific machine cache control CSR (mcache_ctl, 0x7ca) is
+// unknown to the Go assembler, so accesses to it are WORD-encoded below:
+//   csrrs x0, 0x7ca, t0 = 0x7ca2a073
+//   csrrc x0, 0x7ca, t0 = 0x7ca2b073
+#define MCACHE_CTL_IC_EN 0  // bit 0:  instruction cache enable
+#define MCACHE_CTL_DC_EN 16 // bit 16: data cache enable
+
+// DDR controller (offsets relative to DDR_BASE)
+#define DDR_BASE          0x10019000
+#define DDR_STATUS        20  // status register
+#define DDR_STATUS_READY  1   // bit 1: controller ready
+#define DDR_MODE          512
+#define DDR_CTRL          516
+#define DDR_CTRL_INIT     0   // bit 0: trigger initialization
+#define DDR_CTRL_DONE     8   // bit 8: initialization done
+#define DDR_TIMING0       520
+#define DDR_TIMING1       524
+#define DDR_TIMING2       528
+#define DDR_TIMING3       532
+#define DDR_TIMING4       536
+#define DDR_TIMING5       540
+#define DDR_TIMING6       544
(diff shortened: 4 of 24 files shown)


riscv64/irq.s
```diff
@@ -0,0 +1,155 @@
...
+// in a 144-byte 16-byte-aligned frame. WakeG on RISC-V receives the G
+// pointer in T0 rather than A0.
+//
+//go:nosplit
+TEXT ·trapHandler(SB),NOSPLIT|NOFRAME,$0-0
+	// Save RA at the top of the incoming frame, then allocate the frame.
+	// After SUB, 0(SP) holds the saved RA.
+	MOV	X1,  -144(X2)
+	SUB	$144, X2
+	MOV	X5,  8(X2)
```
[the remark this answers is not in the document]
MDr164: I did not yet experiment with TEE, the SoC has some TEE support via (s)PMP but makes sense to look at what you did in GoTEE already to catch any other potential issues.

MDr164:
Don't bother putting time into reviewing this PR until I moved it out of draft. I still need to rework a bunch of the overly verbose code comments.

Edit: I also found a bunch of issues on the actual hardware that were masked by Qemu. This will require a bit of rework and simplification.

MDr164:
I'm not fully happy with NOR boot relying on a flashboot.S file that I have locally as the SoC starts in XIP but the Go binary is linked against DRAM addresses. As we currently don't do position independent firmware I would need to implement some lengthy relocation stub and all that. Any opinion what the best way would be? I kept the DRAM init sequence in case one wants to load the ELF via JTAG; but from NOR flash it requires the boot shim.

abarisani:
> [@MDr164](/MDr164) [MDr164](/MDr164) mentioned this pull request [ Mar 18, 2026 ](#ref-issue-4078467497)
> [ New SoC upstreaming #65 ](/usbarmory/tamago/issues/65)
> Closed
> [@MDr164](/MDr164)
> ###

I don't have a solution for this right now, we relocate in assembly when [loading from MBR in amd64 disks](https://github.com/usbarmory/tamago-example/blob/master/tools/mbr.s).

For other custom boards we have similar assembly helpers at the very beginning and for UEFI we rely on known free memory areas to avoid relocation.

The [loong64 contribution](https://github.com/usbarmory/tamago/issues/70) suggests PIE is possible, so this might develop in something useful.

MDr164:
> The [loong64 contribution](https://github.com/usbarmory/tamago/issues/70) suggests PIE is possible, so this might develop in something useful.

I also looked into that as well and I also believe PIE/PIC is possible for sure but I didn't feel too confident investing into a proper relocation setup. I have a more standalone variant of boot_riscv64.s locally that I just compile with clang/llvm and place into the binary together with the built tamago binary which works just fine but feels a tiny bit hacky. For the NUC980 the ROM is doing all of that essentially and for the other platforms I got here locally up already they also start with working DRAM. Though this platform here jumps into NOR XIP first. I'm fine with keeping the external asm and could upload it here with a mention in the README on how to build it.

abarisani:
> Don't bother putting time into reviewing this PR until I moved it out of draft. I still need to rework a bunch of the overly verbose code comments.
> 
> Edit: I also found a bunch of issues on the actual hardware that were masked by Qemu. This will require a bit of rework and simplification.

Given that the PR is now open shall I take this is ready for review? Thanks

MDr164:
> Given that the PR is now open shall I take this is ready for review? Thanks

Yes, I simplified it a lot and ironed out all the clock/timer issue that where previously masked by Qemu. The code is much less 'controversial' than the NUC980 port I did prior :)

abarisani:
> > Given that the PR is now open shall I take this is ready for review? Thanks
> 
> Yes, I simplified it a lot and ironed out all the clock/timer issue that where previously masked by Qemu. The code is much less 'controversial' than the NUC980 port I did prior :)

Thanks, I'll take a look in the upcoming days. The next 2 weeks will be very busy so please be patient with me, I'll do my best.

MDr164:
There's no urgency with this as there wasn't with the NUC980. I left some of the comments a bit longer to understand where I've taken values from for this board. If the code itself is ok I'm fine with stripping the comments like we did on the previous PR.

There are also several new README.md files so a couple hundred loc are just comments/readmes :)

board/nuclei/eval_soc/timer_riscv64.s
```diff
@@ -0,0 +1,20 @@
...
+// Copyright (c) The TamaGo Authors. All Rights Reserved.
+//
+// Use of this source code is governed by the license
+// that can be found in the LICENSE file.
+
+#include "textflag.h"
+
+// func rdtime() uint64
+//
+// Reads the standard RISC-V time CSR (0xc01) via the rdtime pseudo-instruction
```
abarisani: I think this comment is redundant and can go away.

riscv64/riscv64.go
```diff
@@ -43,11 +43,13 @@ type CPU struct {
 // defined in riscv64.s
 func exit(int32)
 
-// DefaultIdleGovernor is the default CPU idle time management function
+// DefaultIdleGovernor is the default CPU idle time management function.
+// When the scheduler has no work remaining (pollUntil == math.MaxInt64) it
```
abarisani: I assume this doesn't break the existing `TARGET=sifive_u` target of tamago-example? Personally I'd leave the implementation detail outside the comment and ensure that tamago library support is always correctly setup to run the DefaultIdleGovernor.

soc/fisilink/fsl91030/boot_riscv64.s
```diff
@@ -0,0 +1,197 @@
...
+
+//go:build linkcpuinit
+
+#include "textflag.h"
+
+// CSR_MCACHE_CTL (0x7ca) is vendor-specific and unknown to the Go assembler,
+// so it is WORD-encoded:
+//   csrrs x0, 0x7ca, t0 = 0x7ca2a073
+//   csrrc x0, 0x7ca, t0 = 0x7ca2b073
+TEXT cpuinit(SB),NOSPLIT|NOFRAME,$0
```
abarisani, line 82: There is a vast amount of magic constants here, I'd appreciate if they can be converted to (1 << b|1 << c) style so that at least its clear which bits are set (and ideally a comment would show what is set/cleared, see https://github.com/usbarmory/tamago/blob/ea10656381d348058facdde7f3df500d43791e2c/arm64/init.s#L34 for examples).

For other magic constants it would help to clarify their meaning

soc/fisilink/fsl91030/fsl91030.go
```diff
@@ -0,0 +1,143 @@
...
+	WDT_BASE = 0x68000000
+)
+
+// Peripheral instances
+var (
+	// RISC-V core (RV64IMAFDC)
+	RV64 = &riscv64.CPU{
+		Counter:         Counter,
+		TimerMultiplier: 1,
+		// required before Init()
```
abarisani, line 76: This refers to the multiplier and therefore should be moved up?

soc/fisilink/fsl91030/wdt.go
```diff
@@ -0,0 +1,147 @@
...
+	w.Stop()
+
+	var ctrl uint32
+	bits.Set(&ctrl, WDT_CTRL_ENABLE)
+	bits.Set(&ctrl, WDT_CTRL_RSTEN)
+	bits.Set(&ctrl, WDT_CTRL_INTEN)
+	bits.SetN(&ctrl, WDT_CTRL_INTTIME, WDT_CTRL_INTTIME_MASK, uint32(intTimeout))
+	bits.SetN(&ctrl, WDT_CTRL_RSTTIME, WDT_CTRL_RSTTIME_MASK, uint32(rstTimeout))
+
+	w.unlock()
```
abarisani: I'd combine unlock and the subsequent write in a single primitive (e.g. `unlockAndWrite(addr, val uint32)` to ensure this is never missed?

abarisani (changes requested):

soc/fisilink/fsl91030/boot_riscv64.s
MDr164: I added some comments and tried to add explanations where possible. The vendor only provides a single assembly file where this is taken from and several of these regs are not explained there. So for several of these I do not know the meaning either as I do not have access to the NDA'd datasheets.

riscv64/riscv64.go
MDr164: It shouldn't but just to be sure I made that one cou specific now instead of applying the same to every riscv soc.

soc/fisilink/fsl91030/wdt.go
MDr164: Sounds good, so far I didn't run into issue but wrapping it in one call is safer indeed.

abarisani:
I am happy to merge it if you don't have other changes or concerns. Please let me know.

MDr164:
I do have a couple of drivers offline here but those are not needed for booting the platform so I did not put them into this PR. Otherwise I'm running this code here as is locally and it works without any issues so the PR should be good to go from my point of view. Thanks for the review :)

abarisani:
This has now been merged!

A few questions which popped up after the merge.

Do you think the `Setup` write could be moved to `soc/fisilink/fsl91030/fsl91030.go` `Init()` function?

Also does it make sense for `InitUARTPinmux()` to be exported?

MDr164:
The `InitUARTPinmux()` from the soc package is exported as it's being used by the physical board package while the virtual (Qemu) does not need it. Otherwise it could move from the soc to the board package directly. I think what we could do is drop that call from the `Init()` func and keep it as a helper call for platforms needing it?

abarisani:
Are there any side effects to route unconditionally also on QEMU?

MDr164:
Not that I can recall, I think Qemu just no-ops those calls. Let me quickly try to locally call the same on Qemu and see if it behaves weirdly or not.

MDr164:
Seems like it fails with the GPIO pinmux code in Qemu: ` invalid MMIO read at 0x10011048` that's `GPIO_IOF_SEL`. So we still need to keep those calls outside the Qemu package and only in the board one. Though I'm fine with moving it out of the soc package and keep it as an unexported section in the board specific init instead.

PR #69, opened 2026-05-09
add gVNIC datapath and GCE N2D bring-up fixes

marcellmars:
This PR was created across several AI-assisted sessions using Claude Code and Codex.

Builds on #62 by @abarisani.

Adds the GQI-QPL datapath on top of the queue scaffolding from #62: RX descriptor processing with multi-fragment reassembly, TX with QPL FIFO management, ordered Start/Close lifecycle, IRQ unmask after queue creation (Linux `gve_turnup`), device-option parsing, AdminQ status decoding, and the `driver_version` register handshake.

`*GVE` satisfies `usbarmory/go-net.NetworkDevice`:

```go
func (hw *GVE) Receive(buf []byte) (int, error)
func (hw *GVE) Transmit(buf []byte) error
```

## Proposed fixes

Three commits change behaviour in code that already lives in #62 — flagged here for independent review; happy to split, drop, or reshape any of them:

- `fix admin queue counter wrap and empty cmd handling` — `aq.cnt` is now cumulative monotonic uint32 (matching `gve_adminq_kick_cmd`); was wrapped at the ring size, breaking sustained AdminQ traffic past 64 commands. Same commit handles `cmd == nil` for opcodes with no payload (e.g. `DECONFIGURE_DEVICE_RESOURCES`).
- `fix queue setup notification IDs and page list registration` — RX ntfy_id is `1`, TX ntfy_id is `0` (Linux `gve_*_idx_to_ntfy`); previously both used the QPL ID, which collided. `REGISTER_PAGE_LIST` now sends a non-zero `page_size` (added the missing field on the command struct). The page-address list buffer is retained until `UNREGISTER_PAGE_LIST` rather than released immediately.
- `fix counter and IRQ doorbell array allocation` — both regions are now page-aligned with a page-size minimum and zero-cleared before being handed to the device. Previously a few bytes at unspecified alignment were enough on `qemu/microvm` but produced silent dataplane failure on GCE N2D.

## Validation

What I have tested:

- Fork-side unit tests under `kvm/gvnic/`: descriptor layout, AdminQ command serialization, RX reassembly state machine, TX FIFO geometry, `Close` idempotency
- End-to-end bring-up on GCE N2D Confidential VM (SEV-SNP, n2d-standard-2, PCI revision 1):
  - `Init` completes, all 6 init AdminQ commands return `COMMAND_PASSED`
  - DHCP / GCE metadata fetch succeed via the link-local stack
  - TCP through WireGuard works (HTTPS bulk transfer, including YouTube video playback over many minutes)

What I have **not** directly tested in this fork — including which gaps a portdistrict-internal reference implementation (separate from this PR) does cover, with the caveat that the fork's code is not byte-identical and any divergence may behave differently:

- Sustained multi-hour soak under high packet rates. *Reference impl was soak-tested 2 hours under ICMP @ 1 rps; not high-rate.*
- Error/fault injection on the device (`CREATE_*_QUEUE` failures, AdminQ timeouts, descriptor errors mid-stream, partial-init Close). *Reference impl has mock-MMIO tests for these paths; this fork uses `tamago/internal/reg` directly, which has no mock seam, so equivalent tests aren't portable.*
- Multi-queue (single TX + RX queue only)
- PCI revision 0 / legacy `ADMINQ_PFN` path
- MSI-X / interrupt-driven mode (polling-only by design)
- Jumbo frames beyond stock 1460-byte MTU

## Notes

- Polling-only (no MSI-X). `Start()` runs `cleanTxDone` periodically; consumers may also drive a combined RX+poll loop calling `Receive` and `Start1`.
- `Close()` is best-effort — teardown errors are swallowed to allow shutdown to make progress.
- Targets `development`, not `main`, matching #62.


marcellmars:
For reference: how this driver looks from a consumer's side, attached as [nic_gvnic.go](https://github.com/user-attachments/files/27550243/nic_gvnic.go.txt) (a single file from a project which uses this for its exit-node binary, not part of this PR). It drives the proposed API end-to-end on a GCE N2D Confidential VM (SEV-SNP) into a gVisor TCP/IP stack. Posting the snippets 
  that may be useful to you for sanity-checking the API shape; the rest of the file is GCE-specific (link-local bootstrap → metadata → reconfigure)  and orthogonal to the driver.                                                                                                                       
                               
  Lifecycle / construction. The full surface area is pci.Probe + struct literal + Init + MAC + Receive + Transmit + Close:
```go                                                                                                                                                      
  dev := pci.Probe(0, gvnic.PCI_VENDOR, gvnic.PCI_DEVICE)
  if dev == nil {                                                                                                                                     
      return fmt.Errorf("gvnic: no device on PCI bus 0")                                                                                              
  }                                                                                                                                                   
  hw := &gvnic.GVE{                                                                                                                                   
      Device: dev,                                                                                                                                    
      Region: dma.Default(),   
  }                            
  if err := hw.Init(); err != nil {
      return fmt.Errorf("gvnic: init: %w", err)
  }                                                                                                                                                   
  mac := hw.MAC()
```                                                                                                                                                      
 
 
 RX path: combined Receive + cooperative-yield poll loop. TamaGo's scheduler is non-preemptive, so a tight for { Receive(buf) } starves every other goroutine on the runtime (we hit this — runtime.Gosched() after 64 idle iterations is the minimum that keeps the metadata HTTP client + bootstrap server responsive). One goroutine drains both RX and (implicitly) TX completion, since hw.Transmit cleans completions inline:                       
```go                               
  rxBuf := make([]byte, 9216) // jumbo + multi-frag headroom
  go func() {                                                                                                                                         
      idle := 0
      for {                                                                                                                                           
          n, _ := hw.Receive(rxBuf)
          if n > 0 {                                                                                                                                  
              rxInject(link, rxBuf[:n])  // hand to upper stack
              idle = 0                                                                                                                                
              continue         
          }                                                                                                                                           
          idle++               
          if idle >= 64 {      
              runtime.Gosched()
              idle = 0                                                                                                                                
          }
      }                                                                                                                                               
  }()                          
```                               
  TX path: gVisor channel.Endpoint notifier → Transmit. The driver expects a complete Ethernet frame, so the notifier prepends the L2 header from the gVisor egress route. This is the only place the consumer touches L2:
```go                                                                                                                                                      
  type txNotifier struct {     
      hw      *gvnic.GVE       
      link    *channel.Endpoint                                                                                                                       
      localMA tcpip.LinkAddress
  }                                                                                                                                                   
                               
  func (n *txNotifier) WriteNotify() {
      for {
          pkt := n.link.Read()                                                                                                                        
          if pkt == nil {
              return                                                                                                                                  
          }                    
          dstMAC := pkt.EgressRoute.RemoteLinkAddress
          srcMAC := n.localMA
          if len(dstMAC) == 0 {                                                                                                                       
              pkt.DecRef()
              continue                                                                                                                                
          }                    
          proto := uint16(pkt.NetworkProtocolNumber)
          body := pkt.ToView().AsSlice()                                                                                                              
          pkt.DecRef()
                                                                                                                                                      
          frame := make([]byte, 14+len(body))
          copy(frame[0:6], dstMAC)
          copy(frame[6:12], srcMAC)                                                                                                                   
          binary.BigEndian.PutUint16(frame[12:14], proto)
          copy(frame[14:], body)                                                                                                                      
                                                                                                                                                      
          _ = n.hw.Transmit(frame)
      }                                                                                                                                               
  }                            
```                               
  Two observations from running it in-wild that may be worth knowing:                                                                                 
   
  1. Receive ergonomics for cooperative runtimes. A poll-style API is the right shape for TamaGo (no MSI-X, single-threaded model fits), but consumers do need to know that a tight loop will wedge the runtime. Could be worth a one-liner in the package doc on Receive — e.g. "Callers oncooperative-scheduling runtimes should yield (runtime.Gosched) periodically when no frames arrive."                                                 
  2. The dst MAC == 0 branch in the TX notifier is silent. Above, I drop and continue. In production we hit this path during ARP-resolution races and it surfaced as a destination-selective TCP dial failure that took two cycles to diagnose. A return code or a one-time log would have shortened that

marcellmars:
The attached files are part of my workflow with AI assistants and there's some chance a useful informatin could be found there too:
- [tamago-upstream-gvnic.md](https://github.com/user-attachments/files/27550277/tamago-upstream-gvnic.md)
-  [2026-05-09_tamago-upstream-gvnic_retrospective.md](https://github.com/user-attachments/files/27550292/2026-05-09_tamago-upstream-gvnic_retrospective.md)

abarisani:
Hi, as you have seen I was [already working](https://github.com/usbarmory/tamago/pull/62) on a gVNIC implementation and honestly I'd like to finish and carry on with that before evaluating this code.

Thanks

marcellmars:
A mistake.

abarisani:
I've updated https://github.com/usbarmory/tamago/pull/62

IRQ is not fully implemented yet and there are issues to solve under AMD SEV-SNP with SMP, along with some minor polishing.

But apart from this my PR works on sustained traffic under GCP with AMD SEV-SNP and SMP=1.

If you have improvements over my PR feel free to comment, one-by-one, if you have understanding of any potential issues/improvements and their solution.

Thanks

PR #72, opened 2026-06-10
Init features before vtable

MDr164:
Fix GoTEE according to https://github.com/usbarmory/tamago/pull/66#issuecomment-4669837981


MDr164:
@abarisani I assume this should work then?

abarisani:
Yes, this is fine thanks.

Issue #70, opened 2026-06-03
loong64 (LoongArch64) port: working bare-metal PoC, seeking guidance to land it fully


tannevaled:
Hi — I'd like to contribute **`GOOS=tamago GOARCH=loong64`** and have a working proof-of-concept I'd like to align on before pushing toward a complete, upstreamable port.

## What works today (verified on QEMU `-M virt`)

A loong64-patched `tamago-go` + a `loong64` framework package boot a pure-Go program bare-metal and run a **functional Go runtime**:

- boots; **GC/allocator, goroutines + channels, maps, slice growth** all work
- **monotonic `Nanotime`** via the stable timer (`RDTIMED`) + `CPUCFG` frequency
- **RNG** via a timer-seeded PRNG (placeholder for a real entropy source)
- **exception handling**: `CSR.EENTRY` → a handler that reports faults instead of the silent jump-to-0 (verified with a deliberate bad-address read)

Demo output:
```
tamago/loong64: runtime up
monotonic clock: busy-loop took ~ 2529 us
map[5]: 25
PRNG samples (should differ): 41920 57918
goroutine workers total: 150500
tamago/loong64: DONE
triggering a test fault (expect EXC)... EXC
```

## How it's structured (mirrors the riscv64 port)

- **tamago-go fork (small):** `tamago/loong64` added to the GOOS/GOARCH allowlist (`cmd/dist/build.go`, `internal/platform`), `cmd/link/.../loong64` accepts `Htamago`, `os_tamago_loong64.go`, `rt0_tamago_loong64.s`, `sys_tamago_loong64.s` (rt0 bring-up + `GetG` + stubbed `Wake`), `cpu_loong64_other.go` (osInit stub). Upstream Go already ships loong64 codegen + base runtime, so this is glue, not a backend.
- **tamago framework:** `goos/goos_loong64.s` + a `loong64/` package (`Init`/`cpuinit` bring-up — FPU enable via WORD-encoded `csrwr EUEN`, SP setup, `_rt0_tamago_start`; stable-timer counter; minimal exception entry).

Patches available; happy to open a PR in whatever shape you prefer.

## Where I'd like your design guidance (the remaining pieces)

These are coupled to tamago internals / the toolchain, so I want to match your architecture rather than guess:

[...]

PR #73, opened 2026-07-02
add tamago/loong64 support, QEMU virt target

tannevaled:
Adds `GOOS=tamago GOARCH=loong64` (LoongArch 64-bit) support, mirroring the existing riscv64 layering (arch package → SoC drivers → board glue). Follow-up to the compiler-side support in usbarmory/tamago-go#18. Refs #70.

### What's included

- **`loong64/`** — LoongArch 64-bit arch package: CPU bring-up (FPU enable, stack), constant-frequency stable timer (`rdtime.d` + `CPUCFG`), single-entry exception vector with cause/BADV reporting, IRQ enable/disable/wait, `CPUCFG` feature detection, core id. The CPU runs in Direct Address translation mode (no page tables).
- **`soc/loongson/uart`** — NS16550 serial driver; **`soc/loongson/ls3a5000`** — Loongson 3A5000/LS7A SoC glue (console, timer→Nanotime, DRBG RNG, RamStart).
- **`board/qemu/virt`** — QEMU LoongArch `virt` machine board (Hwinit1, Printk, RamSize).
- `internal/reg/reg_loong64.s` and a small `internal/reg/reg8.go` (NS16550 byte registers).

### Assembler note

The Go LoongArch assembler exposes no privileged CSR access, `ertn` or `idle` mnemonics, so those are emitted as hand-encoded `WORD` directives, each verified with `go tool objdump` (e.g. `csrwr R4, csr = 0x04000000 | (csr<<10) | 0x24`).

### Validation

Built and run end-to-end under `qemu-system-loongarch64 -M virt` (v11.0.1):

```
qemu-system-loongarch64 -M virt -m 256M -nographic -kernel <elf>
```

Verified: boot + serial console, goroutine scheduling, GC, accurate `time.Sleep` (20 ms → ~21 ms), `crypto/rand`, and clean exception reporting on a deliberate fault (`ecode 8 ADE`, correct `badv`, Go stack trace). Existing riscv64/amd64/arm64/arm boards still build; `go vet` and `gofmt` clean.

Unlike the riscv64 `qemu` targets no BIOS is required — the QEMU `virt` machine loads the ELF and jumps to its entry directly (the low 2 MiB is reserved for boot info + device tree, so the text segment is linked above it, e.g. `-T 0x1000000`).

### Scope

Primary hardware target is Loongson 3A5000/3A6000 + LS7A; QEMU `virt` is the reference target validated here. A dedicated real-silicon board (with cached DMW windows, real memory map) is a natural follow-up — QEMU `virt` is the only LoongArch target testable in emulation, so real-HW specifics are intentionally out of scope for this PR. Happy to adjust SoC/package naming to your preference.

### DIFF
--- a/board/qemu/virt/README.md
+++ b/board/qemu/virt/README.md
@@ -1,0 +1,117 @@
+TamaGo - bare metal Go - QEMU virt support
+==========================================
+
+tamago | https://github.com/usbarmory/tamago  
+
+Copyright (c) The TamaGo Authors. All Rights Reserved.  
+
+![TamaGo gopher](https://github.com/usbarmory/tamago/wiki/images/tamago.svg?sanitize=true)
+
+Authors
+=======
+
+tannevaled  
+https://github.com/tannevaled  
+
+Introduction
+============
+
+TamaGo is a framework that enables compilation and execution of unencumbered Go
+applications on bare metal processors.
+
+The `virt` package provides support for the QEMU `virt` machine, currently
+targeting [LoongArch](https://www.qemu.org/docs/master/system/loongarch/virt.html)
+(`GOARCH=loong64`) configured with a single core.
+
+Documentation
+=============
+
+[![Go Reference](https://pkg.go.dev/badge/github.com/usbarmory/tamago.svg)](https://pkg.go.dev/github.com/usbarmory/tamago)
+
+For more information about TamaGo see its
+[repository](https://github.com/usbarmory/tamago) and
+[project wiki](https://github.com/usbarmory/tamago/wiki).
+
+For the underlying driver support for this board see package
+[ls3a5000](https://github.com/usbarmory/tamago/tree/master/soc/loongson/ls3a5000).
+
+The package API documentation can be found on
+[pkg.go.dev](https://pkg.go.dev/github.com/usbarmory/tamago).
+
+Supported hardware
+==================
+
+| SoC               | Board                                                                          | SoC package                                                                        | Board package                                                                |
+|-------------------|--------------------------------------------------------------------------------|------------------------------------------------------------------------------------|------------------------------------------------------------------------------|
+| Loongson 3A5000   | [QEMU virt](https://www.qemu.org/docs/master/system/loongarch/virt.html)       | [ls3a5000](https://github.com/usbarmory/tamago/tree/master/soc/loongson/ls3a5000)  | [qemu/virt](https://github.com/usbarmory/tamago/tree/master/board/qemu/virt) |
+
+Compiling
+=========
+
+Go applications are required to set `GOOSPKG` to the desired
+[runtime/goos](https://github.com/usbarmory/tamago-go/tree/latest/src/runtime/goos)
+overlay and import the relevant board package to ensure that hardware
+initialization and runtime support take place:
+
+```golang
+import (
+	_ "github.com/usbarmory/tamago/board/qemu/virt"
+)
+```
+
+Go applications can be compiled as usual, using the compiler built for the
+TamaGo framework, but with the addition of the following flags/variables:
+
+```sh
+GOOS=tamago GOOSPKG=github.com/usbarmory/tamago GOARCH=loong64 \
+	${TAMAGO} build -ldflags "-T 0x1000000 -R 0x1000" main.go
+```
+
+The QEMU `virt` machine reserves the low 2 MiB of RAM for boot information and
+the device tree, therefore the text segment must be linked above that region
+(the example above links at 16 MiB).
+
+Build tags
+==========
+
+The following build tags allow applications to override the package own
+definition for the `runtime/goos` overlay:
+
+* `linkramsize`: exclude `ramSize` from `mem.go`
+* `linkprintk`: exclude `printk` from `console.go`
+
+Executing and debugging
+=======================
+
+The target can be executed under emulation as follows:
+
+```sh
+qemu-system-loongarch64 \
+	-machine virt -m 256M \
+	-nographic -monitor none -serial stdio -net none \
+	-kernel main
+```
+
+Unlike the riscv64 `qemu` targets no BIOS is required, as the QEMU LoongArch
+`virt` machine loads the ELF image and jumps to its entry point directly.
+
+The emulated target can be debugged with GDB by adding the `-S -s` flags to the
+previous execution command, this will make qemu wait for a GDB connection that
+can be launched as follows:
+
+```sh
+loongarch64-unknown-linux-gnu-gdb -ex "target remote 127.0.0.1:1234" main
+```
+
+License
+=======
+
+tamago | https://github.com/usbarmory/tamago  
+Copyright (c) The TamaGo Authors. All Rights Reserved.
+
+These source files are distributed under the BSD-style license found in the
+[LICENSE](https://github.com/usbarmory/tamago/blob/master/LICENSE) file.
+
+The TamaGo logo is adapted from the Go gopher designed by Renee French and
+licensed under the Creative Commons 3.0 Attributions license. Go Gopher vector
+illustration by Hugo Arganda.
--- a/loong64/README.md
+++ b/loong64/README.md
@@ -1,0 +1,67 @@
+TamaGo - bare metal Go - LoongArch 64-bit support
+=================================================
+
+tamago | https://github.com/usbarmory/tamago  
+
+Copyright (c) The TamaGo Authors. All Rights Reserved.  
+
+![TamaGo gopher](https://github.com/usbarmory/tamago/wiki/images/tamago.svg?sanitize=true)
+
+Authors
+=======
+
+tannevaled  
+https://github.com/tannevaled  
+
+Introduction
+============
+
+TamaGo is a framework that enables compilation and execution of unencumbered Go
+applications on bare metal processors.
+
+The [loong64](https://github.com/usbarmory/tamago/tree/master/loong64) package
+provides support for LoongArch 64-bit CPUs.
+
+The CPU runs in Direct Address translation mode (`CRMD.DA`), therefore no page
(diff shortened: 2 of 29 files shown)


board/qemu/virt/README.md
```diff
@@ -0,0 +1,120 @@
...
+tamago | https://github.com/usbarmory/tamago  
+
+Copyright (c) The TamaGo Authors. All Rights Reserved.  
+
+![TamaGo gopher](https://github.com/usbarmory/tamago/wiki/images/tamago.svg?sanitize=true)
+
+Authors
+=======
+
+Andrea Barisani
```
abarisani: We didn't author this, change as appropriate. Thanks.

loong64/README.md
```diff
@@ -0,0 +1,70 @@
...
+tamago | https://github.com/usbarmory/tamago  
+
+Copyright (c) The TamaGo Authors. All Rights Reserved.  
+
+![TamaGo gopher](https://github.com/usbarmory/tamago/wiki/images/tamago.svg?sanitize=true)
+
+Authors
+=======
+
+Andrea Barisani
```
abarisani: Same comment as earlier, credit actual author please.

soc/loongson/ls3a5000/README.md
```diff
@@ -0,0 +1,72 @@
...
+tamago | https://github.com/usbarmory/tamago  
+
+Copyright (c) The TamaGo Authors. All Rights Reserved.  
+
+![TamaGo gopher](https://github.com/usbarmory/tamago/wiki/images/tamago.svg?sanitize=true)
+
+Authors
+=======
+
+Andrea Barisani
```
abarisani: Change authors.

soc/loongson/uart/uart.go
```diff
@@ -0,0 +1,94 @@
...
+		hw.Tx(c)
+	}
+
+	return len(buf), nil
+}
+
+// Read receives the available bytes from the serial port into the argument
+// buffer.
+func (hw *UART) Read(buf []byte) (n int, _ error) {
+	for n = 0; n < len(buf); n++ {
```
abarisani: I think this should be fixed for an off by one just like https://github.com/usbarmory/tamago/commit/671275bdb2fa7503b3cdef96e9a6f3a253a13988 ?

loong64/loong64.go
```diff
@@ -0,0 +1,65 @@
...
+
+// This package supports 64-bit cores.
+const XLEN = 64
+
+// CPU instance
+type CPU struct {
+	sync.Mutex
+
+	// Counter represents the function to obtain the system counter for
+	// operation of [CPU.GetTime] and [CPU.SetTime].
```
abarisani: The fact that `Rdtime` is used as default, when not set, should be documented.

abarisani (changes requested):

board/qemu/virt/README.md
tannevaled: Fixed — the Authors section now credits the actual author (tannevaled).

loong64/README.md
tannevaled: Fixed — credited tannevaled.

soc/loongson/ls3a5000/README.md
tannevaled: Fixed — credited tannevaled.

soc/loongson/uart/uart.go
tannevaled: Good catch — aligned `Read`/`Write` with the canonical implementation from #71 (loop now increments only after a successful `Rx`, so the returned count is correct); `Write` already returned `len(buf), nil`.

loong64/loong64.go
tannevaled: Documented — the `Counter` field comment now notes it defaults to `Rdtime` at `CPU.Init` time when left nil.

tannevaled:
Pushed `106ca76` addressing all review comments:

- Authors sections now credit the actual author (tannevaled) across the three READMEs.
- `soc/loongson/uart`: aligned `Read`/`Write` with the #71 canonical implementation.
- `loong64`: documented that `Counter` defaults to `Rdtime` at `CPU.Init`.

Rebuilt and re-booted under `qemu-system-loongarch64 -M virt` (boot + console still good); `go vet`/`gofmt` clean; other arch boards still build. Thanks for the review.

tannevaled:
Follow-up on the `uart` review: while verifying the `Read` change under QEMU with real serial input (via a `-serial pipe:` fifo) I found and fixed a related bug — `Init` was writing `FCR=0x01` (FIFO enable only, without the RX/TX FIFO reset bits), so the RX FIFO didn't buffer and bursts were lost to overrun (a 4-byte burst read back as 1 byte). Changed to `FCR=0xC7` (enable + clear RX/TX FIFOs + 14-byte trigger); a burst now reads back correctly (`n`=3–4, count logic verified: reads all available bytes, empty read returns 0). Raw LSR/RBR probing confirmed the register model. Folded into `3f8d3f3`.

tannevaled:
Added interrupt servicing to the `loong64` package (`35393cd`), mirroring the riscv64/arm64 model:

- `trapHandler` now dispatches on `ESTAT.Ecode` — interrupts (Ecode 0) go to `handleInterrupt`, faults to `systemException`.
- `handleInterrupt` saves/restores the full GPR set, relays to `os/signal` (`IRQ_SIGNAL`), masks `PRMD.PIE` on return and `ertn`s back to the interrupted context.
- `CPU.ServiceInterrupts`, `EnableInterrupt`/`DisableInterrupt` (ECFG.LIE), and constant-timer helpers (`SetTimer`/`ClearTimerInterrupt` via TCFG/TICLR).

Verified under QEMU with a periodic 5 ms timer interrupt: ~49 interrupts serviced over 300 ms, the interrupted goroutine resumes correctly (time keeps advancing), and the exception path still reports faults. `go vet`/`gofmt` clean; other arch boards still build.

(Every hand-encoded CSR `WORD` was objdump-verified — which caught a real bug: an early `SAVE0` stash was mis-encoded as a write to the read-only `CPUID` CSR, spotted via the QEMU `BADI` trace.)

abarisani:
Given that the `virt` machine exists in qemu for various architectures I'd propose the following renaming:

`console.go` > `console_loong64.go`
`virt.go`    > `virt_loong64.go`

While `mem.go` can remain as it's not specific to loong64 support, its header needs to be changed to reflect generic QEMU `virt` machine support along with the README.

loong64/irq.go
```diff
@@ -0,0 +1,74 @@
...
+
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, IRQ_SIGNAL)
+
+	for {
+		// To avoid losing interrupts, re-enabling must happen only after
+		// we are waiting.
+		go cpu.EnableInterrupts()
+		<-c
+		isr()
```
abarisani, line 80: How is the ISR supposed to get which interrupt fired? The API for reading it should be exported.

abarisani:
I adapted https://github.com/usbarmory/tamago-example to this target (I will push it once this is merged) and everything is in order, I have minor style fixes that I can take or after this is merged.

I left one comment in the review to be addressed and also please resolve the reg8.go conflict.

Thanks!

abarisani:
I am testing interrupts as follows:

```go
ls3a5000.LA64.ServiceInterrupts(func() {
    print("irq!")
})
```

But this hangs my qemu, am I missing something? Thanks.

tannevaled:
Pushed `51dc567` addressing all comments (rebased onto `development`, so the `reg8.go` conflict is resolved — `Read8`/`Write8` now live in `internal/reg/reg_loong64.s` per the new convention).

**Board file renaming** (as requested):
- `console.go` → `console_loong64.go`
- `virt.go` → `virt_loong64.go`
- `mem.go` kept, header changed to generic *QEMU virt support*; README title/intro likewise generalized to the QEMU `virt` machine (currently `GOARCH=loong64`).

**"How does the ISR know which interrupt fired?"** — exported `CPU.InterruptStatus() uint64`, returning the `ESTAT.IS` pending-line bitmap:

```go
cpu.ServiceInterrupts(func() {
    if cpu.InterruptStatus()&(1<<loong64.TimerInterrupt) != 0 {
        cpu.ClearTimerInterrupt()
    }
})
```

**The QEMU hang** — your snippet doesn't arm an interrupt source, so `ServiceInterrupts` blocks, the runtime goes idle and the CPU halts (I reproduced it: a single `EXCP_HLT`). A source needs to be armed and the ISR must acknowledge it. Full working example, verified under `qemu-system-loongarch64 -M virt`:

```go
cpu := ls3a5000.LA64
go cpu.ServiceInterrupts(func() {
    if cpu.InterruptStatus()&(1<<loong64.TimerInterrupt) != 0 {
        cpu.ClearTimerInterrupt()
    }
})
cpu.SetTimer(int64(5*time.Millisecond), true) // periodic
cpu.EnableInterrupt(loong64.TimerInterrupt)   // unmask the timer line in ECFG.LIE
```

This services the timer interrupt continuously while the interrupted goroutines keep running (time advances). `go vet`/`gofmt` clean; riscv64/amd64/arm64/arm boards still build.

Glad tamago-example already runs on this target — happy to fold in your style fixes here or after merge, your call. Thanks!

abarisani:
Merged, thanks for this contribution!

My cleanup commit is https://github.com/usbarmory/tamago/commit/91e08435ca1a555df018a322d7512b97599c5bc7

I will push tamago-example changes later today.

abarisani:
Actually the tamago-example patch will need to wait for https://go-review.googlesource.com/c/sys/+/788141

abarisani:
Never mind, I worked around the dependency issue at https://github.com/usbarmory/tamago-example/commit/4055aa4080c277a9cb79aa497ef7078d8150e3e0

PR #77, opened 2026-07-29
lan969x: add temperature sensor support

lperkov:
Configure the on-die sensor for continuous sampling and expose the latest valid temperature through the LAN969x singleton. Derive the sensor timing from the 328.125 MHz system clock and use the conversion documented in datasheet.


abarisani:
I merged development and modified the driver style to match the rest of the codebase, please test.

Invocation now is:

```go
lan969x.TEMP.Init()
c, valid := lan969x.TEMP.Read()
```

lperkov:
Tested it now with this last patch it works fine.

PR #78, opened 2026-07-30
internal/reg: fix Get16

lperkov:
Get16 modifies the register by setting the requested bit and returns no value, duplicating Set16 instead of matching Get and Get64.

Read the bit and return its state without modifying the register.


abarisani:
Thanks for catching this!

PR #80, opened 2026-07-30
flexcom: add buffered interrupt-driven USART receive

lperkov:
This series corrects FLEXCOM write-only register access and improves the USART receive path when interrupts and hardware FIFOs are used.

### DIFF
--- a/internal/reg/reg8.go
+++ b/internal/reg/reg8.go
@@ -1,0 +1,14 @@
+// https://github.com/usbarmory/tamago
+//
+// Copyright (c) The TamaGo Authors. All Rights Reserved.
+//
+// Use of this source code is governed by the license
+// that can be found in the LICENSE file.
+
+package reg
+
+// Read8 reads an 8-bit register.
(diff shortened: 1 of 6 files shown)


abarisani:
I think buffering should be implemented outside the tamago library in application defined ISRs.

If this is doable, not to pollute the UART driver with buffering or further interrupt awareness, please refactor the PR just to include any eventual bugs on the current implementation and leave anything that can be handled externally by the application to external ISRs, thanks.

lperkov:
Kept only the command-register fix and hardware FIFO support. Software buffering is no longer part of this PR.

abarisani:
@lperkov FYI with https://github.com/usbarmory/tamago/commit/fb68df3f4edcf69cc8eed2ada91e29b6a481ad1c  I aligned `(*FLEXCOM).Write` to all other implementations, this should mean no change. I also aligned amd64 `(*UART).Rx` implementation to this refactor (which feels better).

PR #84, opened 2026-07-30
lan969x: add watchdog timer support

lperkov:
LAN969x integrates a Synopsys DesignWare APB watchdog at 0xe0090000. Applications currently need to duplicate its register map and use raw MMIO to configure resets, restart the counter, and inspect its state.

Add a serialized watchdog driver and expose the LAN969x instance. Support direct-reset and interrupt-first modes while keeping periodic kicking and reboot lifecycle policy in the application. The fixed timeout ranges use the verified 250 MHz LAN969x fabric clock.

Live LAN969x registers identify component type 0x44570120, version 0x3131302a, and parameters 0x100d0e42. Tested both modes on LAN969x. A ten-second CPU halt asserted the first-stage interrupt and restarted the counter; resuming the kicker cleared it. Direct mode reset the SoC.


abarisani:
@lperkov for single register reads the mutex is overkill, I cleaned this up (and improved documentation) in https://github.com/usbarmory/tamago/commit/4b278a40e30ef199fac2c9339b28ba2437672a71

PR #87, opened 2026-08-04
improve Microchip GPIO abstraction

abarisani:
This PR improves the Microchip GPIO driver abstraction to make it more compliant with similar implementations within the code base.


abarisani:
@lperkov can you please test this change?

lperkov:
@abarisani I tested it and it works.

PR #88, opened 2026-08-05
move Ethernet PHY drivers to individual sub-packages

abarisani:
This PR moves all PHY configuration and control code to its own sub-packages under `phy`.


abarisani:
@lperkov please test, once merged this will allow you to integrate a new PHY driver if desired.

lperkov:
@abarisani I tested it and it works. Thanks!

PR #90, opened 2026-08-06
lan8804: add PHY driver

lperkov:
LAN8804 extended-page accesses require a dependent Clause 22 sequence, but TamaGo has no driver to keep that sequence atomic or report reset and link state consistently.

Add a bus-independent driver under the common PHY hierarchy. Implement the shared PHY interface and provide serialized direct and Extended Page access, reset, identifier, link, speed, and duplex status.

### DIFF
--- a/phy/microchip/lan8804/phy.go
+++ b/phy/microchip/lan8804/phy.go
@@ -140,28 +140,6 @@ func (hw *PHY) WriteExtendedRegister(page int, address uint16, data uint16) (err
 	return hw.write(EP_ACCESS_DATA, data)
 }
 
-// ModifyExtendedRegister updates selected bits in an Extended Page register.
-func (hw *PHY) ModifyExtendedRegister(page int, address uint16, mask uint16, set uint16) (err error) {
-	if hw.miim == nil {
-		return errors.New("invalid PHY instance")
-	}
-
-	if err = hw.selectExtended(page, address); err != nil {
-		return
-	}
-
-	var data uint16
-	if data, err = hw.read(EP_ACCESS_DATA); err != nil {
-		return
-	}
-
-	if err = hw.selectExtended(page, address); err != nil {
-		return
-	}
-
-	return hw.write(EP_ACCESS_DATA, (data&^mask)|(set&mask))
-}
-
 // Reset performs a software hard reset and waits for its completion.
 func (hw *PHY) Reset() (err error) {
 	if hw.miim == nil {
@@ -174,8 +152,8 @@ func (hw *PHY) Reset() (err error) {
 
 	deadline := time.Now().Add(hw.Timeout)
 
+	var control uint16
 	for {
-		var control uint16
 		if control, err = hw.read(BASIC_CONTROL); err != nil {
 			return
 		}
--- a/phy/microchip/lan8804/phy.go
+++ b/phy/microchip/lan8804/phy.go
@@ -1,0 +1,303 @@
+// LAN8804/LAN8814 Ethernet PHY support
+// https://github.com/usbarmory/tamago
+//
+// Copyright (c) The TamaGo Authors. All Rights Reserved.
+//
+// Use of this source code is governed by the license
+// that can be found in the LICENSE file.
+
+// Package lan8804 implements a driver for the Microchip LAN8804/LAN8814
+// Gigabit Ethernet PHYs adopting the following reference specifications:
+//   - Microchip - LAN8804D/LAN8814D GPHY Register Definitions - DS00004286D
+//   - Microchip - LAN8804D Datasheet - DS00003591K
+package lan8804
+
+import (
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/usbarmory/tamago/phy"
+)
+
+const (
+	BASIC_CONTROL  = 0x00
+	BASIC_STATUS   = 0x01
+	PHY_ID_1       = 0x02
+	PHY_ID_2       = 0x03
+	EP_ACCESS_CTRL = 0x16
+	EP_ACCESS_DATA = 0x17
+	CONTROL        = 0x1f
+
+	CTRL_RESET        = 15
+	CTRL_ANEG_ENABLE  = 12
+	CTRL_ANEG_RESTART = 9
+
+	STATUS_LINK          = 2
+	STATUS_ANEG_COMPLETE = 5
+
+	EP_FUNCTION = 14
+	EP_INDEX    = 0
+
+	EP_ADDRESS = 0
+	EP_DATA    = 1
+
+	CONTROL_SPEED_1000 = 6
+	CONTROL_SPEED_100  = 5
+	CONTROL_SPEED_10   = 4
+	CONTROL_DUPLEX     = 3
+)
+
+// Timeout is the default timeout for PHY operations.
+const Timeout = 100 * time.Millisecond
+
+// Status represents the resolved PHY link state.
+type Status struct {
+	Link                    bool
+	AutoNegotiationComplete bool
+	Speed                   int
+	FullDuplex              bool
+}
+
+// PHY represents a LAN8804/LAN8814 PHY port.
+type PHY struct {
+	// Timeout for PHY operations
+	Timeout time.Duration
+
+	miim  phy.MIIM
+	pa    int
+	speed int
+}
+
+// Init initializes a LAN8804/LAN8814 PHY instance for register access.
+func (hw *PHY) Init(addr int, miim phy.MIIM) (err error) {
+	if miim == nil {
+		return errors.New("invalid PHY instance")
+	}
+
+	if addr < 0 || addr > 0x1f {
+		return errors.New("invalid PHY address")
+	}
+
+	hw.miim = miim
+	hw.pa = addr
+	hw.speed = 0
+
+	if hw.Timeout == 0 {
+		hw.Timeout = Timeout
+	}
+
+	return
+}
+
+func (hw *PHY) read(address int) (data uint16, err error) {
+	return hw.miim.ReadPHYRegister(hw.pa, address)
+}
+
+func (hw *PHY) write(address int, data uint16) (err error) {
+	return hw.miim.WritePHYRegister(hw.pa, address, data)
+}
+
+func (hw *PHY) selectExtended(page int, address uint16) (err error) {
+	if page < 0 || page > 0x1f {
+		return errors.New("invalid LAN8804 extended page")
+	}
+
+	if err = hw.write(EP_ACCESS_CTRL, uint16(EP_ADDRESS<<EP_FUNCTION|page<<EP_INDEX)); err != nil {
+		return
+	}
+
+	if err = hw.write(EP_ACCESS_DATA, address); err != nil {
+		return
+	}
+
+	return hw.write(EP_ACCESS_CTRL, uint16(EP_DATA<<EP_FUNCTION|page<<EP_INDEX))
+}
+
+// ReadExtendedRegister reads a vendor-specific Extended Page register.
+func (hw *PHY) ReadExtendedRegister(page int, address uint16) (data uint16, err error) {
+	if hw.miim == nil {
+		return 0, errors.New("invalid PHY instance")
+	}
+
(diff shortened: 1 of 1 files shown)


phy/microchip/lan8804/phy.go
```diff
@@ -0,0 +1,410 @@
...
+	}
+
+	hw.speed = status.Speed
+
+	return
+}
+
+// Address returns the PHY address passed at [PHY.Init].
+func (hw *PHY) Address() (addr int) {
+	hw.Lock()
```
abarisani: Not necessary here.

phy/microchip/lan8804/phy.go
```diff
@@ -0,0 +1,410 @@
...
+	defer hw.Unlock()
+
+	return hw.pa
+}
+
+// Speed returns the last resolved PHY link speed.
+func (hw *PHY) Speed() (speed int) {
+	hw.Lock()
```
abarisani: Not necessary here.

phy/microchip/lan8804/phy.go
```diff
@@ -0,0 +1,410 @@
...
+	up = basic&(1<<STATUS_LINK) != 0
+	return
+}
+
+// Status returns link, auto-negotiation, speed, and duplex state.
+func (hw *PHY) Status() (status Status, err error) {
+	hw.Lock()
+	defer hw.Unlock()
+
+	if err = hw.initialized(); err != nil {
```
abarisani: We keep the same LOC (minus `initialized()`) bu keeping the same check and returned error within the functions. I'd remove `initialized` for this reason.

phy/microchip/lan8804/phy.go
```diff
@@ -0,0 +1,410 @@
...
+		return
+	}
+
+	control |= (1 << CTRL_ANEG_ENABLE) | (1 << CTRL_ANEG_RESTART)
+	return hw.write(BASIC_CONTROL, control)
+}
+
+// Identifier returns the PHY identifier registers as one 32-bit value.
+func (hw *PHY) Identifier() (id uint32, err error) {
+	hw.Lock()
```
abarisani: Not necessary here.

phy/microchip/lan8804/phy.go
```diff
@@ -0,0 +1,410 @@
...
+func (hw *PHY) read(address int) (data uint16, err error) {
+	return hw.miim.ReadPHYRegister(hw.pa, address)
+}
+
+func (hw *PHY) write(address int, data uint16) (err error) {
+	return hw.miim.WritePHYRegister(hw.pa, address, data)
+}
+
+// ReadRegister reads a direct PHY register.
+func (hw *PHY) ReadRegister(address int) (data uint16, err error) {
```
abarisani: This is not used within the driver and merely a proxy to existing MIIM functions, I'd use struct embedding to define these functions for all PHYs if this can be a common pattern.

phy/microchip/lan8804/phy.go
```diff
@@ -0,0 +1,410 @@
...
+
+	if err = validateRegister(address); err != nil {
+		return
+	}
+
+	return hw.read(address)
+}
+
+// WriteRegister writes a direct PHY register.
+func (hw *PHY) WriteRegister(address int, data uint16) (err error) {
```
abarisani: This is not used within the driver and merely a proxy to existing MIIM functions, I'd use struct embedding to define these functions for all PHYs if this can be a common pattern.

phy/microchip/lan8804/phy.go
```diff
@@ -0,0 +1,410 @@
...
+
+	if err = validateRegister(address); err != nil {
+		return
+	}
+
+	return hw.write(address, data)
+}
+
+// ModifyRegister updates selected bits in a direct PHY register.
+func (hw *PHY) ModifyRegister(address int, mask uint16, set uint16) (err error) {
```
abarisani: This is not used within the driver and merely a proxy to existing MIIM functions, I'd use struct embedding to define these functions for all PHYs if this can be a common pattern.

phy/microchip/lan8804/phy.go
```diff
@@ -0,0 +1,410 @@
...
+
+	var data uint16
+	if data, err = hw.read(address); err != nil {
+		return
+	}
+
+	return hw.write(address, (data&^mask)|(set&mask))
+}
+
+func (hw *PHY) selectExtended(page int, address uint16) (err error) {
```
abarisani, line 101: This is not used within the driver and merely a proxy to existing MIIM functions, I'd use struct embedding to define these functions for all PHYs if this can be a common pattern.

phy/microchip/lan8804/phy.go
```diff
@@ -0,0 +1,410 @@
...
+
+	if err = hw.write(EP_ACCESS_DATA, address); err != nil {
+		return
+	}
+
+	return hw.write(EP_ACCESS_CTRL, uint16(EP_DATA<<EP_FUNCTION|page<<EP_INDEX))
+}
+
+// ReadExtendedRegister reads a vendor-specific Extended Page register.
+func (hw *PHY) ReadExtendedRegister(page int, address uint16) (data uint16, err error) {
```
abarisani, line 118: This is not used within the driver and merely a proxy to existing MIIM functions, I'd use struct embedding to define these functions for all PHYs if this can be a common pattern.

phy/microchip/lan8804/phy.go
```diff
@@ -0,0 +1,410 @@
...
+
+	if err = hw.selectExtended(page, address); err != nil {
+		return
+	}
+
+	return hw.read(EP_ACCESS_DATA)
+}
+
+// WriteExtendedRegister writes a vendor-specific Extended Page register.
+func (hw *PHY) WriteExtendedRegister(page int, address uint16, data uint16) (err error) {
```
abarisani, line 131: This is not used within the driver and merely a proxy to existing MIIM functions, I'd use struct embedding to define these functions for all PHYs if this can be a common pattern.

phy/microchip/lan8804/phy.go
```diff
@@ -0,0 +1,410 @@
...
+
+	if err = hw.selectExtended(page, address); err != nil {
+		return
+	}
+
+	return hw.write(EP_ACCESS_DATA, data)
+}
+
+// ModifyExtendedRegister updates selected bits in an Extended Page register.
+func (hw *PHY) ModifyExtendedRegister(page int, address uint16, mask uint16, set uint16) (err error) {
```
abarisani: This is not used within the driver and merely a proxy to existing MIIM functions, I'd use struct embedding to define these functions for all PHYs if this can be a common pattern.

phy/microchip/lan8804/phy.go
lperkov: Removed the lock from `Address`.

phy/microchip/lan8804/phy.go
lperkov: Removed the lock from `Speed`.

phy/microchip/lan8804/phy.go
lperkov: Removed `initialized`; each operation now keeps the check and error inline.

phy/microchip/lan8804/phy.go
lperkov: Removed the lock from `Identifier`.

phy/microchip/lan8804/phy.go
lperkov: Removed this direct-register proxy. Callers can use `phy.MIIM` for ordinary Clause 22 access.

phy/microchip/lan8804/phy.go
lperkov: Removed this direct-register proxy. Callers can use `phy.MIIM` for ordinary Clause 22 access.

phy/microchip/lan8804/phy.go
lperkov: Removed this direct-register proxy. Callers can use `phy.MIIM` for ordinary Clause 22 access.

phy/microchip/lan8804/phy.go
lperkov: Kept this helper because the Extended Page methods use it for the LAN8804-specific page/address/data selection sequence; it is not a direct MIIM proxy.

phy/microchip/lan8804/phy.go
lperkov: Kept this operation because `phy.MIIM` exposes only direct Clause 22 registers. LAN8804 Extended Page access requires the vendor-specific dependent sequence, so it remains in this driver.

phy/microchip/lan8804/phy.go
lperkov: Kept this operation because `phy.MIIM` exposes only direct Clause 22 registers. LAN8804 Extended Page access requires the vendor-specific dependent sequence, so it remains in this driver.

phy/microchip/lan8804/phy.go
lperkov: Kept this operation because `phy.MIIM` exposes only direct Clause 22 registers. LAN8804 Extended Page access requires the vendor-specific dependent sequence, so it remains in this driver.

lperkov:
@abarisani can you please review again?

phy/microchip/lan8804/phy.go
```diff
@@ -0,0 +1,303 @@
...
```
abarisani: I think Read/Write can be wrapped by callers rather than having this 3rd method here?

phy/microchip/lan8804/phy.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+
+	if err = hw.write(BASIC_CONTROL, 1<<CTRL_RESET); err != nil {
+		return
+	}
+
+	deadline := time.Now().Add(hw.Timeout)
+
+	for {
+		var control uint16
```
abarisani: Move outside loop.

phy/microchip/lan8804/phy.go
```diff
@@ -0,0 +1,303 @@
...
+	if low, err = hw.read(PHY_ID_2); err != nil {
+		return
+	}
+
+	id = uint32(high)<<16 | uint32(low)
+	return
+}
+
+func (hw *PHY) link() (basic uint16, err error) {
+	if _, err = hw.read(BASIC_STATUS); err != nil {
```
abarisani, line 211: Why 2 reads?

phy/microchip/lan8804/phy.go
```diff
@@ -0,0 +1,303 @@
...
+}
+
+// Status returns link, auto-negotiation, speed, and duplex state.
+func (hw *PHY) Status() (status Status, err error) {
+	if hw.miim == nil {
+		return status, errors.New("invalid PHY instance")
+	}
+
+	var basic, control uint16
+	if basic, err = hw.link(); err != nil {
```
abarisani, line 241: Given that `link()` is always used to get a boolean with `STATUS_LINK` maybe link() should simply be `func link() bool` ?

phy/microchip/lan8804/phy.go
lperkov: Done.

phy/microchip/lan8804/phy.go
lperkov: Done.

phy/microchip/lan8804/phy.go
lperkov: BASIC_STATUS.STATUS_LINK is latch-low. The first read clears any latched link-down state; the second reports the current link state.

phy/microchip/lan8804/phy.go
lperkov: Status() also uses STATUS_ANEG_COMPLETE from the same Basic Status value, so returning only a link boolean would require another register read. I kept the register value; renaming the helper to basicStatus() would make that contract clearer.

phy/microchip/lan8804/phy.go
abarisani: This should be a comment in the code then as it's not quite obvious.

lperkov: Done.

PR #91, opened 2026-08-06
fan: add LAN969x controller support

lperkov:


soc/microchip/fan/fan.go
```diff
@@ -0,0 +1,177 @@
...
+package fan
+
+import (
+	"sync"
+
+	"github.com/usbarmory/tamago/bits"
+	"github.com/usbarmory/tamago/internal/reg"
+)
+
+const (
```
abarisani, line 25: Use newlines to separate groups of registers+offsets among each others as well as constant values, consistently with the rest of tamago code base.

soc/microchip/fan/fan.go
```diff
@@ -0,0 +1,177 @@
...
+	// PulsesPerSecond updates the counter once per second with the number of
+	// tachometer pulses received during the previous second.
+	PulsesPerSecond CounterMode = iota
+	// TotalPulses configures the counter as a wrapping total of tachometer
+	// pulses.
+	TotalPulses
+)
+
+// Config defines fan controller output and tachometer behavior.
+type Config struct {
```
abarisani, line 61: Individual fields needs documentation.

soc/microchip/lan969x/README.md
```diff
@@ -42,7 +42,7 @@ Supported hardware
 
 | SoC               | Related board packages                                                                             | Peripheral drivers                                                                                                                                                                         |
 |-------------------|----------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
```
abarisani: Align the table header too.

soc/microchip/fan/fan.go
```diff
@@ -0,0 +1,177 @@
...
+	return
+}
+
+// TachoCount returns the raw 16-bit tachometer count and whether the controller
+// is initialized.
+func (hw *FAN) TachoCount() (count uint16, valid bool) {
+	hw.Lock()
+	defer hw.Unlock()
+
+	if !hw.run {
```
abarisani: The controller might have been initialized previously by a bootloader, I'd just check `hw.Base` instead of run, get rid of the mutex check, and simply return the count read. It feels more useful this way.

soc/microchip/fan/fan.go
```diff
@@ -0,0 +1,177 @@
...
+	return
+}
+
+// Frequency returns the configured PWM frequency and whether the controller is
+// initialized.
+func (hw *FAN) Frequency() (hz uint32, valid bool) {
+	hw.Lock()
+	defer hw.Unlock()
+
+	if !hw.run {
```
abarisani: The controller might have been initialized previously by a bootloader, I'd just check hw.Base instead of run, get rid of the mutex check, and simply return the frequency read. It feels more useful this way.

soc/microchip/fan/fan.go
```diff
@@ -0,0 +1,177 @@
...
+	reg.SetN(hw.Base+FAN_CFG, CFG_DUTY_CYCLE, CFG_DUTY_CYCLE_MASK, uint32(duty))
+}
+
+// Duty returns the configured PWM duty cycle and whether the controller is
+// initialized.
+func (hw *FAN) Duty() (duty uint8, valid bool) {
+	hw.Lock()
+	defer hw.Unlock()
+
+	if !hw.run {
```
abarisani: The controller might have been initialized previously by a bootloader, I'd just check hw.Base instead of run, get rid of the mutex check, and simply return the duty cycle read. It feels more useful this way.

soc/microchip/fan/fan.go
```diff
@@ -0,0 +1,177 @@
...
+// SetDuty sets the PWM duty cycle, where 0 is always off and 255 is always on.
+func (hw *FAN) SetDuty(duty uint8) {
+	hw.Lock()
+	defer hw.Unlock()
+
+	if !hw.run {
+		panic("initialize fan controller")
+	}
+
+	reg.SetN(hw.Base+FAN_CFG, CFG_DUTY_CYCLE, CFG_DUTY_CYCLE_MASK, uint32(duty))
```
abarisani, line 120: If there are no adverse effects I'd remove mutex and hw.run checks here as this is a single write.

soc/microchip/fan/fan.go
```diff
@@ -0,0 +1,177 @@
...
+	reg.Write(hw.Base+PWM_FREQ, frequency)
+
+	var cfg uint32
+	bits.SetTo(&cfg, CFG_INV_POL, config.Inverted)
+	bits.SetTo(&cfg, CFG_PWM_OPEN_COL_ENA, config.OpenCollector)
+	bits.SetTo(&cfg, CFG_GATE_ENA, config.GateTacho)
+	bits.SetTo(&cfg, CFG_FAN_STAT_CFG, config.CounterMode == TotalPulses)
+	reg.Write(hw.Base+FAN_CFG, cfg)
+
+	hw.run = true
```
abarisani: Per my other comments consider removing this.

soc/microchip/fan/fan.go
lperkov: Done. Duty now checks Base, performs no locking, and returns the configured duty cycle directly.

soc/microchip/lan969x/README.md
lperkov: Done. The header and separator are aligned with the expanded row.

soc/microchip/fan/fan.go
lperkov: Done. Each Config field now documents its effect.

soc/microchip/fan/fan.go
lperkov: Done. TachoCount now checks Base, performs no locking, and returns the register value directly.

soc/microchip/fan/fan.go
lperkov: Done. The run state was removed; accessors use Base and read the hardware state directly.

soc/microchip/fan/fan.go
lperkov: Done. Frequency now checks Base, performs no locking, and returns the configured frequency directly.

soc/microchip/fan/fan.go
lperkov: Done. The register fields, register groups, and clock constants are now separated consistently with the existing drivers.

soc/microchip/fan/fan.go
lperkov: Done. SetDuty now performs the register update without the mutex or run check.

lperkov:
@abarisani can you please review again?

PR #96, opened 2026-08-10
add idle time reporting

abarisani:
This patch adds idle time reporting for single-core platforms, to aid `top` like reporting.


abarisani:
@lperkov this should help you to account for CPU usage on ARM64

PR #97, opened 2026-08-11
flexcom: avoid deferred unlock during early init

lperkov:
InitUSART runs from Hwinit1 on the runtime system stack. Debug builds with optimization disabled lower its deferred unlock to runtime.deferprocStack, which aborts with "defer on system stack" before the console banner.

Unlock explicitly so early initialization behaves independently of compiler optimization. A JTAG-staged debug runtime reported its exact identity on LAN969x hardware after this change.

### DIFF
--- a/soc/microchip/flexcom/flexcom.go
+++ b/soc/microchip/flexcom/flexcom.go
@@ -59,9 +59,9 @@ func (hw *FLEXCOM) mode(n int) {
 // mode, this is mutually exclusive with other modes.
 func (hw *FLEXCOM) InitUSART() {
 	hw.Lock()
-	defer hw.Unlock()
 
 	if hw.Base == 0 {
+		hw.Unlock()
 		panic("invalid FLEXCOM controller instance")
 	}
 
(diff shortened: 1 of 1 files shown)


soc/microchip/flexcom/flexcom.go
```diff
@@ -59,9 +59,9 @@ func (hw *FLEXCOM) mode(n int) {
 // mode, this is mutually exclusive with other modes.
 func (hw *FLEXCOM) InitUSART() {
 	hw.Lock()
-	defer hw.Unlock()
 
 	if hw.Base == 0 {
+		hw.Unlock()
```
abarisani, line 64: If we panic this is pointless.

soc/microchip/flexcom/flexcom.go
```diff
@@ -82,6 +82,9 @@ func (hw *FLEXCOM) InitUSART() {
 	hw.USART.fmr = hw.Base + FLEX_USART_OFFSET + FLEX_US_FMR
 
 	hw.USART.init()
+
+	// Hwinit1 runs on the system stack.
```
abarisani, line 86: Better comment would be: "avoid defer as we might be called from system stack".

PR #104, opened 2026-09-02
soc/microchip: add LAN969x UVOV diagnostics

lperkov:


soc/microchip/uvov/uvov.go
```diff
@@ -0,0 +1,216 @@
...
+package uvov
+
+import (
+	"errors"
+
+	"github.com/usbarmory/tamago/bits"
+	"github.com/usbarmory/tamago/internal/reg"
+)
+
+// Register definitions follow Microchip's BSD-3-Clause LAN969x support at
```
abarisani: Replace with `UVOV registers`, consistently with other drivers, as we don't want to maintain links here if we can avoid it.

soc/microchip/uvov/uvov.go
```diff
@@ -0,0 +1,216 @@
...
+// https://github.com/microchip-ung/arm-trusted-firmware/tree/master/plat/microchip/lan969x.
+const (
```
abarisani, line 25: Follow existing style which groups registers/offsets together, separating them from each other.

soc/microchip/uvov/uvov.go
```diff
@@ -0,0 +1,216 @@
...
+	return false
+}
+
+// UVOV represents one under-voltage and over-voltage monitor instance.
+type UVOV struct {
+	// Base is the register base.
+	Base uint32
+}
+
+var railLayout = [railCount]struct {
```
abarisani: Avoid this static variable, move it to a UVOV instance, use `Init` if required.

abarisani (changes requested):

soc/microchip/uvov/uvov.go
```diff
@@ -0,0 +1,237 @@
...
+	InterruptStatus uint32
+	InterruptEnable uint32
+	UnderEnabled    bool
+	OverEnabled     bool
+	Trim            TrimSnapshot
+	Rails           [railCount]RailSnapshot
+}
+
+// LiveFault reports whether a detector currently reports an invalid voltage.
+func (snapshot Snapshot) LiveFault() bool {
```
abarisani, line 131: Should be all pointer methods?

soc/microchip/uvov/uvov.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// Snapshot reads the complete monitor configuration and status without
+// changing detector, interrupt, or reset state.
+func (hw *UVOV) Snapshot() (snapshot Snapshot, err error) {
+	if hw.Base == 0 {
+		return snapshot, errors.New("invalid UVOV instance")
+	}
+
+	rails := [railCount]railRegisters{
```
abarisani, line 189: Are these specific to Microchip LAN969x or common to other SoCs? If not generic should they be received as argument by Snapshot and all methods be converted to a single iteration rather than looping rails in all methods?

PR #106, opened 2026-09-03
soc/microchip: support eMMC on LAN969x

lperkov:
Add an SDHCI package for eMMC and bind the LAN969x SDMMC0 controller to it.

The API initializes and detects sector-addressed eMMC, reports card identity and capacity, reads Extended CSD, transfers aligned blocks, and synchronizes enabled device caches.

The initial scope excludes SD cards, DMA, multi-block writes, erase and trim, eMMC partition switching, boot partitions, RPMB, high-speed timing, HS200, and HS400.

### DIFF
--- a/soc/microchip/sdhci/adma.go
+++ b/soc/microchip/sdhci/adma.go
@@ -1,0 +1,50 @@
+// Microchip Secure Digital Host Controller Interface support
+// https://github.com/usbarmory/tamago
+//
+// Copyright (c) The TamaGo Authors. All Rights Reserved.
+//
+// Use of this source code is governed by the license
+// that can be found in the LICENSE file.
+
+package sdhci
+
+import "encoding/binary"
+
+const (
+	// ADMA2 data and descriptor addresses are word aligned.
+	admaAlignment = 4
+
+	admaDescriptorSize = 8
+	// Largest non-zero, word-aligned value in the 16-bit length field.
+	admaMaxLength = 65532
+
+	admaValid    = 1 << 0
+	admaEnd      = 1 << 1
+	admaTransfer = 0b10 << 4
+)
+
+func admaTableSize(size int) int {
+	return (size + admaMaxLength - 1) / admaMaxLength * admaDescriptorSize
+}
+
+func admaTable(address uint, size int) []byte {
+	table := make([]byte, admaTableSize(size))
+
+	for offset := 0; size > 0; offset += admaDescriptorSize {
+		length := min(size, admaMaxLength)
+		attribute := uint16(admaValid | admaTransfer)
+
+		if length == size {
+			attribute |= admaEnd
+		}
+
+		binary.LittleEndian.PutUint16(table[offset:], attribute)
+		binary.LittleEndian.PutUint16(table[offset+2:], uint16(length))
+		binary.LittleEndian.PutUint32(table[offset+4:], uint32(address))
+
+		address += uint(length)
+		size -= length
+	}
+
+	return table
+}
--- a/soc/microchip/sdhci/cmd.go
+++ b/soc/microchip/sdhci/cmd.go
@@ -1,0 +1,272 @@
+// Microchip Secure Digital Host Controller Interface support
+// https://github.com/usbarmory/tamago
+//
+// Copyright (c) The TamaGo Authors. All Rights Reserved.
+//
+// Use of this source code is governed by the license
+// that can be found in the LICENSE file.
+
+package sdhci
+
+import (
+	"fmt"
+	"runtime"
+	"time"
+
+	"github.com/usbarmory/tamago/bits"
+	"github.com/usbarmory/tamago/internal/reg"
+)
+
+var (
+	// CommandInhibitTimeout controls how long a command waits for the host
+	// command and data paths to become idle.
+	CommandInhibitTimeout = 200 * time.Millisecond
+	// CommandTimeout controls how long a command waits for completion.
+	CommandTimeout = 500 * time.Millisecond
+)
+
+type cmdParams struct {
+	responseType uint16
+	commandType  uint16
+	openDrain    bool
+	indexCheck   bool
+	crcCheck     bool
+	dataPresent  bool
+}
+
+var cmds = map[uint16]cmdParams{
+	// CMD0 - GO_IDLE_STATE - reset card
+	0: {responseType: CR_RESPTYP_NORESP, openDrain: true},
+	// CMD1 - SEND_OP_COND - send operating conditions
+	1: {responseType: CR_RESPTYP_RL48, openDrain: true},
+	// CMD2 - ALL_SEND_CID - get unique card identification
+	2: {responseType: CR_RESPTYP_RL136, openDrain: true, crcCheck: true},
+	// CMD3 - SET_RELATIVE_ADDR - set relative card address
+	3: {responseType: CR_RESPTYP_RL48, openDrain: true, indexCheck: true, crcCheck: true},
+	// CMD6 - SWITCH - switch mode of operation
+	6: {responseType: CR_RESPTYP_RL48BSY, indexCheck: true, crcCheck: true},
+	// CMD7 - SELECT/DESELECT_CARD - enter or leave transfer state
+	7: {responseType: CR_RESPTYP_RL48BSY, indexCheck: true, crcCheck: true},
+	// CMD8 - SEND_EXT_CSD - read extended device data
+	8: {responseType: CR_RESPTYP_RL48, indexCheck: true, crcCheck: true, dataPresent: true},
+	// CMD9 - SEND_CSD - read device data
+	9: {responseType: CR_RESPTYP_RL136, crcCheck: true},
+	// CMD12 - STOP_TRANSMISSION - stop a multi-block transfer
+	12: {responseType: CR_RESPTYP_RL48BSY, commandType: CR_CMDTYP_ABORT, crcCheck: true},
+	// CMD13 - SEND_STATUS - read card status
+	13: {responseType: CR_RESPTYP_RL48, indexCheck: true, crcCheck: true},
+	// CMD17 - READ_SINGLE_BLOCK - read one block
+	17: {responseType: CR_RESPTYP_RL48, indexCheck: true, crcCheck: true, dataPresent: true},
+	// CMD18 - READ_MULTIPLE_BLOCK - read consecutive blocks
+	18: {responseType: CR_RESPTYP_RL48, indexCheck: true, crcCheck: true, dataPresent: true},
+	// CMD24 - WRITE_BLOCK - write one block
+	24: {responseType: CR_RESPTYP_RL48, indexCheck: true, crcCheck: true, dataPresent: true},
+}
+
+func commandValue(index uint16, params cmdParams) (command uint16) {
+	bits.SetN16(&command, CR_RESPTYP, CR_RESPTYP_MASK, params.responseType)
+	bits.SetN16(&command, CR_CMDTYP, CR_CMDTYP_MASK, params.commandType)
+	bits.SetTo16(&command, CR_CMDICEN, params.indexCheck)
+	bits.SetTo16(&command, CR_CMDCCEN, params.crcCheck)
+	bits.SetTo16(&command, CR_DPSEL, params.dataPresent)
+	bits.SetN16(&command, CR_CMDIDX, CR_CMDIDX_MASK, index)
+
+	return
+}
+
+func (hw *SDHCI) cmd(index uint16, argument uint32) (uint32, error) {
+	value, _, _, err := hw.runCommand(index, argument, 0)
+
+	return value, err
+}
+
+func (hw *SDHCI) runCommand(index uint16, argument uint32, ignoredErrors uint16) (value uint32, ignored uint16, issued bool, err error) {
+	params, ok := cmds[index]
+	if !ok {
+		return 0, 0, false, fmt.Errorf("sdhci: unsupported command CMD%d", index)
+	}
+
+	// clear pending status
+	reg.Write16(hw.eistr, allInterrupts)
+	reg.Write16(hw.nistr, allInterrupts)
+
+	usesDataLine := params.responseType == CR_RESPTYP_RL48BSY || params.dataPresent
+
+	var inhibitMask uint32
+	bits.Set(&inhibitMask, PSR_CMDINHC)
+
+	if params.commandType != CR_CMDTYP_ABORT && usesDataLine {
+		bits.Set(&inhibitMask, PSR_CMDINHD)
+	}
+
+	if !reg.WaitFor(CommandInhibitTimeout, hw.psr, 0, int(inhibitMask), 0) {
+		var resetMask uint16
+		bits.Set16(&resetMask, SRR_SWRSTCMD)
+
+		if usesDataLine {
+			bits.Set16(&resetMask, SRR_SWRSTDAT)
+		}
+
+		if resetErr := hw.reset(uint8(resetMask), ControllerSetupTimeout); resetErr != nil {
+			return 0, 0, false, fmt.Errorf("sdhci: command inhibit timeout (recovery failed: %v)", resetErr)
+		}
+
+		return 0, 0, false, fmt.Errorf("CMD%d command inhibit timeout", index)
+	}
+
+	// configure command signaling
+	mode := uint16(reg.Read8(hw.mc1r))
+	bits.SetN16(&mode, MC1R_CMDTYP, MC1R_CMDTYP_MASK, 0)
+	bits.SetTo16(&mode, MC1R_OPD, params.openDrain)
+	bits.Set16(&mode, MC1R_FCD)
+	reg.Write8(hw.mc1r, uint8(mode))
+
+	// issue card command
+	reg.Write(hw.arg1r, argument)
+	reg.Write16(hw.cr, commandValue(index, params))
+
+	value, ignored, err = hw.pollStatusIgnoring(1<<NISTR_CMDC, CommandTimeout, ignoredErrors)
+	if err != nil {
+		var resetMask uint16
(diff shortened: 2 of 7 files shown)


soc/microchip/sdhci/cmd.go
```diff
@@ -0,0 +1,239 @@
...
+import (
+	"fmt"
+	"runtime"
+	"time"
+
+	"github.com/usbarmory/tamago/internal/reg"
+)
+
+const (
+	commandInhibitTimeout = 200 * time.Millisecond
```
abarisani: Export these and add documentation so that users can tune as required.

soc/microchip/sdhci/cmd.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+
+		if resetErr := hw.reset(mask, controllerSetupTimeout); resetErr != nil {
+			return 0, 0, false, fmt.Errorf("sdhci: command inhibit timeout (recovery failed: %v)", resetErr)
+		}
+
+		return 0, 0, false, fmt.Errorf("CMD%d command inhibit timeout", index)
+	}
+
+	mc1r := reg.Read8(hw.mc1r) &^ (MC1R_CMDTYP_MASK<<MC1R_CMDTYP | 1<<MC1R_OPD)
```
abarisani: I feel this entire file could be more consistent and readbale using `bits` package like most of tamago does?

soc/microchip/sdhci/cmd.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+func (hw *SDHCI) stopTransmission(transferErr error) error {
+	status, ignored, _, stopErr := hw.runCommand(12, 0, EISTR_DAT_LINE_ERROR_MASK)
+
+	if stopErr == nil && ignored != 0 {
+		stopErr = hw.reset(1<<SRR_SWRSTDAT, controllerSetupTimeout)
+	}
+
+	if stopErr == nil {
```
abarisani: I'd keep a more traditional logic flow for all conditionals, example:

```
if err = checkR1(status); err != nil {
    return trasnferErr // (or err directly?)
}
```

soc/microchip/sdhci/mmc.go
```diff
@@ -0,0 +1,285 @@
...
+
+	if err != nil {
+		return
+	}
+
+	if err = checkR1(status); err != nil {
+		return
+	}
+
+	err = hw.waitState(CURRENT_STATE_TRAN, timeout)
```
abarisani: return this directly

soc/microchip/sdhci/mmc.go
```diff
@@ -0,0 +1,285 @@
...
+	hostControl := reg.Read8(hw.hc1r)
+	hostControl &^= 1 << HC1R_DW_4BIT
+	hostControl |= 1 << HC1R_EXTDW
+	reg.Write8(hw.hc1r, hostControl)
+
+	if err = hw.setClockFrequency(mmcLegacyClockHz); err != nil {
+		return
+	}
+
+	err = hw.detectCapabilitiesMMC()
```
abarisani: return this directly

soc/microchip/sdhci/mmc.go
```diff
@@ -0,0 +1,285 @@
...
+
+	hw.ready = true
+
+	return
+}
+
+// Ready reports whether Detect completed and no transfer failure has forced
+// controller reinitialization.
+func (hw *SDHCI) Ready() bool {
+	hw.Lock()
```
abarisani: Lock/Unlock is unnecessary here

soc/microchip/sdhci/mmc.go
```diff
@@ -0,0 +1,285 @@
...
+	defer hw.Unlock()
+
+	return hw.ready
+}
+
+// Info returns detected card metadata.
+func (hw *SDHCI) Info() CardInfo {
+	hw.Lock()
```
abarisani: Lock/Unlock is unnecessary here

soc/microchip/sdhci/sdhci.go
```diff
@@ -0,0 +1,662 @@
...
+	GCK_PRESCALER_MASK = 0xff
+)
+
+// SDHCI constants
+const (
+	// data transfer direction
+	WRITE = 0
+	READ  = 1
+
+	controllerSetupTimeout = 500 * time.Millisecond
```
abarisani: Consider which timeouts should be exported and documented for users to tune as needed.

soc/microchip/sdhci/sdhci.go
```diff
@@ -0,0 +1,662 @@
...
+func (hw *SDHCI) enableGenericClock(prescaler uint32) error {
+	value := reg.Read(hw.GCK)
+
+	if value&1<<GCK_ENA != 0 {
+		// A previous firmware stage may leave SDCLK running. Stop it only
+		// after the command and data paths become idle, while the inherited
+		// functional clock is still available.
+		inhibitMask := 1<<PSR_CMDINHC | 1<<PSR_CMDINHD
+		if !reg.WaitFor(controllerSetupTimeout, hw.psr, 0, inhibitMask, 0) {
+			return fmt.Errorf("sdhci: inherited controller busy gck=0x%08x psr=0x%08x ccr=0x%04x", value, reg.Read(hw.psr), reg.Read16(hw.ccr))
```
abarisani: Most error lines are quite verbose throughout the package. Consider simplifying them and add a single `dumpRegisters` method which provides a string to append to such errors logs as needed. (see line 153 of cmd.go for another example). Also always ponder what is useful for the user, are these register dumps really useful for the user?

soc/microchip/sdhci/sdhci.go
```diff
@@ -0,0 +1,662 @@
...
+
+	inhibitMask := 1<<PSR_CMDINHC | 1<<PSR_CMDINHD
+	if !reg.WaitFor(clockSetupTimeout, hw.psr, 0, inhibitMask, 0) {
+		return errors.New("sdhci: clock change timeout waiting for idle")
+	}
+
+	reg.Clear16(hw.ccr, CCR_SDCLKEN)
+
+	var clock uint16
+	clock |= 1 << CCR_INTCLKEN
```
abarisani: Use bits package.

soc/microchip/sdhci/sdhci.go
```diff
@@ -0,0 +1,662 @@
...
+	if err := hw.enableGenericClock(prescaler); err != nil {
+		return err
+	}
+
+	if err := hw.reset(1<<SRR_SWRSTALL, controllerSetupTimeout); err != nil {
+		return fmt.Errorf("%w gck=0x%08x ccr=0x%04x nistr=0x%04x eistr=0x%04x", err, reg.Read(hw.GCK), reg.Read16(hw.ccr), reg.Read16(hw.nistr), reg.Read16(hw.eistr))
+	}
+
+	reg.Write8(hw.tcr, dataTimeoutCounter)
+	reg.Write8(hw.pcr, PCR_SDBVSEL_3V3<<PCR_SDBVSEL|1<<PCR_SDBPWR)
```
abarisani: add one-liner comments here and throught register writes that are not immediately clear what they do (example https://github.com/usbarmory/tamago/blob/7af6ac9043d441c691d1b0e0e543ce451aa0cc27/soc/nxp/usdhc/sd.go#L190) so that we have a clue on what is going on and why.

soc/microchip/sdhci/sdhci.go
```diff
@@ -0,0 +1,662 @@
...
+	GCK uint32
+	// Generic Clock source frequency in Hz
+	ParentClock uint32
+	// Requested Generic Clock frequency in Hz
+	TargetClock uint32
+	// Controller pin configuration, skipped when nil
+	ConfigurePins func() error
+	// DMA memory used for ADMA2 descriptors and data. It must be accessible
+	// by the controller, non-cacheable, and below 4 GiB.
+	DMA *dma.Region
```
abarisani: Why do you need a DMA region to be passed externally and not use dma package native functions which use the Default region?

soc/microchip/sdhci/sdhci.go
```diff
@@ -0,0 +1,662 @@
...
+
+	// LAN969x implements the SDHCI ADMA2 interface. Require the advertised
+	// capability, then select its 32-bit descriptor format because DMA memory
+	// is constrained below 4 GiB.
+	if !reg.Get(hw.capr, CAPR_ADMA2) {
+		return errors.New("sdhci: controller does not support ADMA2")
+	}
+
+	hostControl := reg.Read8(hw.hc1r)
+	hostControl &^= HC1R_DMASEL_MASK << HC1R_DMASEL
```
abarisani: use bits package

soc/microchip/sdhci/sdhci.go
```diff
@@ -0,0 +1,662 @@
...
+func (hw *SDHCI) transferDMA(index uint16, direction uint32, lba uint32, buf []byte, blocks uint16) (err error) {
+	dmaAddress, _ := hw.DMA.Reserve(len(buf), admaAlignment)
+	defer hw.DMA.Release(dmaAddress)
+
+	if direction == WRITE {
+		writeDMAWords(dmaAddress, buf)
+	}
+
+	descriptors := admaTable(dmaAddress, len(buf))
+	descriptorAddress, _ := hw.DMA.Reserve(len(descriptors), admaAlignment)
```
abarisani: If there is a requirement for 32-bit aligned access then dma is not the rigth tool for the job and it brings no benefit whatsoever to the code, in fact `writeDMAWords` and `readDMAWords` in the end do not use the buffer that it allocated, but merely the address itself.

If reserved byte arrays can be used as they are for `copy` then use those, if as the comment suggest only 32-bit aligned access can be performed then do not use the dma package.

abarisani (changes requested):

soc/microchip/sdhci/adma.go
```diff
@@ -0,0 +1,50 @@
...
+	admaEnd      = 1 << 1
+	admaTransfer = 0b10 << 4
+)
+
+func admaTableSize(size int) int {
+	return (size + admaMaxLength - 1) / admaMaxLength * admaDescriptorSize
+}
+
+func admaTable(address uint, size int) []byte {
+	table := make([]byte, admaTableSize(size))
```
abarisani: In line with the rest of the tamago code base, and to be more Go idiomatic, make this a struct and then when needed serialize it with a [Bytes() function](https://github.com/usbarmory/tamago/blob/7af6ac9043d441c691d1b0e0e543ce451aa0cc27/kvm/sev/vmsa.go#L146).

soc/microchip/sdhci/cmd.go
```diff
@@ -0,0 +1,272 @@
...
+		if status&expected != 0 {
+			reg.Write16(hw.nistr, expected)
+			return reg.Read(hw.rr), ignored, nil
+		}
+
+		if time.Now().After(deadline) {
+			return 0, ignored, fmt.Errorf("sdhci: status 0x%04x timeout (%s)", expected, hw.dumpRegisters())
+		}
+
+		runtime.Gosched()
```
abarisani: I'd remove this and re-introduce it only if it turns out to be an issue.

soc/microchip/sdhci/cmd.go
```diff
@@ -0,0 +1,272 @@
...
+
+		if bits.Get(&status, STATUS_READY_FOR_DATA) && bits.GetN(&status, STATUS_CURRENT_STATE, STATUS_CURRENT_STATE_MASK) == uint32(state) {
+			return nil
+		}
+
+		if time.Now().After(deadline) {
+			return fmt.Errorf("sdhci: card ready timeout status=0x%08x", status)
+		}
+
+		runtime.Gosched()
```
abarisani: I'd remove this and re-introduce it only if it turns out to be an issue.

soc/microchip/sdhci/mmc.go
```diff
@@ -0,0 +1,279 @@
...
+		// CMD1 - SEND_OP_COND - send operating conditions
+		response, err := hw.cmd(1, arg)
+
+		if err == nil && bits.GetN(&response, MMC_OCR_BUSY, 1) == 1 {
+			hw.card.OCR = response
+			ready = bits.GetN(&response, MMC_OCR_ACCESS_MODE, 0b11) == ACCESS_MODE_SECTOR
+			break
+		}
+
+		runtime.Gosched()
```
abarisani: I'd remove this and re-introduce it only if it turns out to be an issue.

soc/microchip/sdhci/mmc.go
```diff
@@ -0,0 +1,279 @@
...
+	time.Sleep(MMCPowerUpDelay)
+	start := time.Now()
+
+	for time.Since(start) <= MMCDetectTimeout {
+		// CMD1 - SEND_OP_COND - send operating conditions
+		response, err := hw.cmd(1, arg)
+
+		if err == nil && bits.GetN(&response, MMC_OCR_BUSY, 1) == 1 {
+			hw.card.OCR = response
+			ready = bits.GetN(&response, MMC_OCR_ACCESS_MODE, 0b11) == ACCESS_MODE_SECTOR
```
abarisani: return the value directly

soc/microchip/sdhci/mmc.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+
+		return errors.New("CMD1 SEND_OP_COND: card did not power up")
+	}
+
+	if err = hw.initMMC(); err != nil {
+		return
+	}
+
+	hw.ready = true
```
abarisani: I'd put this as `} else { ..` in the defer so that you can then `return hw.initMMC()` directly.

soc/microchip/sdhci/sdhci.go
abarisani: This remains unanswered, if there is no real use case for this I'd drop the field and always use the default region. If there is value in having it please rename it as `Region` consistently with the rest of tamago drivers.

soc/microchip/sdhci/sdhci.go
```diff
@@ -0,0 +1,717 @@
...
+	controllerReady bool
+	ready           bool
+
+	// detected card properties
+	card CardInfo
+
+	maxBlocks int
+}
+
+func (hw *SDHCI) dumpRegisters() string {
```
abarisani: If the driver is working and tested consider removing this and making errors less verbose.

soc/microchip/sdhci/sdhci.go
```diff
@@ -0,0 +1,717 @@
...
+	if !reg.WaitFor8(timeout, hw.srr, 0, int(mask), 0) {
+		return fmt.Errorf("sdhci: controller reset 0x%02x timeout (%s)", mask, hw.dumpRegisters())
+	}
+
+	return nil
+}
+
+func (hw *SDHCI) initController(prescaler uint32) error {
+	if hw.ConfigurePins != nil {
+		if err := hw.ConfigurePins(); err != nil {
```
abarisani: I see no value in having this function invocation within the driver, given that it's invoked with no arguments or fields from it.

If it can be invoked by the application before `(*SHDCI).Init` I'd simply do so. For instance other boards configure USDHC pins [on their own](https://github.com/usbarmory/tamago/blob/master/board/usbarmory/mk2/usdhc.go).

soc/microchip/sdhci/sdhci.go
```diff
@@ -0,0 +1,717 @@
...
+// Init initializes the controller. Detect must be called afterward to
+// initialize the eMMC card.
+func (hw *SDHCI) Init() error {
+	hw.Lock()
+	defer hw.Unlock()
+
+	hw.controllerReady = false
+	hw.ready = false
+	hw.card = CardInfo{}
+	hw.maxBlocks = 0
```
abarisani: Unnecessary as it gets re-assigned.

soc/microchip/sdhci/sdhci.go
```diff
@@ -0,0 +1,717 @@
...
+
+	switch dtd {
+	case WRITE:
+		for len(buf) > 0 {
+			if err = hw.writeBlock(index, uint32(lba), buf[:BlockSize]); err != nil {
+				return
+			}
+
+			if err = hw.waitState(CURRENT_STATE_TRAN, WriteTimeout); err != nil {
+				err = hw.invalidateTransfer(err)
```
abarisani: return this directly

soc/microchip/sdhci/sdhci.go
```diff
@@ -0,0 +1,717 @@
...
+
+			if err = hw.readBlocks(index, uint32(lba), buf[:length], uint16(blocks)); err != nil {
+				return
+			}
+
+			buf = buf[length:]
+			lba += blocks
+		}
+	default:
+		err = errors.New("sdhci: invalid transfer direction")
```
abarisani: return this directly

abarisani (changes requested):

soc/microchip/sdhci/mmc.go
```diff
@@ -58,7 +57,7 @@ var (
 	MMCPowerUpDelay = 10 * time.Millisecond
 )
 
-func (hw *SDHCI) voltageValidationMMC() (ready bool) {
+func (hw *SDHCI) voltageValidationMMC() bool {
```
abarisani: Keep the named variable to document its meaning.

soc/microchip/sdhci/sdhci.go
```diff
@@ -448,15 +424,15 @@ func (hw *SDHCI) Init() error {
...
 	}
 
-	if hw.DMA == nil {
-		hw.DMA = dma.Default()
+	if hw.Region == nil {
+		hw.Region = dma.Default()
 	}
 
-	if hw.DMA == nil {
+	if hw.Region == nil {
```
abarisani: Unreachable because you override it if nil in the earlier statement, remove.

abarisani (changes requested):

PR #108, opened 2026-09-04
lan969x: expose explicit reset scopes

lperkov:
Separate the system, VCore, and ARM Core 0 warm-reset mechanisms so callers can select the intended propagation.

### DIFF
--- a/soc/microchip/lan969x/reset.go
+++ b/soc/microchip/lan969x/reset.go
@@ -13,16 +13,64 @@ import (
 )
 
 const (
+	// CPU registers
+	CPU_RESET           = CPU_BASE + 0x84
+	CPU_RESET_PROT_STAT = CPU_BASE + 0x88
+
+	// GCB registers
 	GCB_SOFT_RST = GCB_BASE + 0x0c
 	SOFT_SWC_RST = 1
 )
 
-// Reset asserts the switch core reset field causing the SoC to restart with a
-// soft reset.
-//
-// Note that only the SoC itself is guaranteed to restart as, depending on the
-// board hardware layout, the system might remain powered (which might not be
-// desirable). See respective board packages for cold reset options.
-func Reset() {
-	reg.Set(GCB_SOFT_RST, SOFT_SWC_RST)
+// ResetTarget specifies which parts of the chip Reset resets.
+type ResetTarget uint8
+
+const (
+	// Core0ColdReset stops the primary ARM core and its debug interface.
+	// Execution does not resume automatically and requires external recovery.
+	Core0ColdReset ResetTarget = 1 << iota
+	// WatchdogReset holds the watchdog in reset until a later Reset call clears it.
+	WatchdogReset
+	// DDRReset holds DDR memory in reset and stops programs that depend on it.
+	DDRReset
+	// L2CacheReset resets the ARM L2 cache and clears automatically when complete.
+	L2CacheReset
+	// JTAGReset holds the ARM JTAG controller in reset until a later Reset call clears it.
+	JTAGReset
+	// ProcessorDebugReset holds the ARM debug components in reset until a later Reset call clears it.
+	ProcessorDebugReset
+	// Core0WarmReset restarts the primary ARM core without a full processor reboot.
+	Core0WarmReset
+	// VCoreReset performs a full processor reboot without restarting the entire chip.
+	VCoreReset
+)
+
+// ResetProtection specifies which parts of the chip are protected from reset.
+type ResetProtection uint8
+
+const (
+	// PCIeProtection keeps the PCIe controller running during a VCore reset.
+	PCIeProtection ResetProtection = 1 << iota
+	// WatchdogResetStatus reports that a watchdog timeout reset VCore. It is not a protection setting.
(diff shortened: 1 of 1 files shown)


soc/microchip/lan969x/reset.go
```diff
@@ -13,16 +13,42 @@ import (
...
+// Callers must disable frame DMA before invoking this function. The board and
+// external devices remain powered; see board packages for cold reset options.
+func ResetVCore() {
+	reg.Write(CPU_RESET, 1<<RESET_VCORE_RST)
+
+	// wait for reset
+	select {}
+}
+
+// ResetCore0Warm performs a warm reset of ARM Core 0 while the VCore system
```
abarisani: Simplify this description, make documentation clear to users and a reset should be a simple matter.

soc/microchip/lan969x/reset.go
```diff
@@ -13,16 +13,42 @@ import (
...
+// ResetSystem initiates a system soft reset. The switch core always resets;
+// the VCore also resets unless protected through CPU_RESET_PROT_STAT.
+// Callers must disable frame DMA before resetting an unprotected VCore.
+func ResetSystem() {
 	reg.Set(GCB_SOFT_RST, SOFT_SWC_RST)
 }
+
+// ResetVCore restarts the VCore CPU from boot ROM by asserting a VCore reset.
+//
+// Callers must disable frame DMA before invoking this function. The board and
```
abarisani: Simplify this description, make documentation clear to users and a reset should be a simple matter. From a user standpoint I have no idea what "frame DMA" is and how to disable it, if there is a call out there to do so then refer it.

soc/microchip/lan969x/reset.go
```diff
@@ -13,16 +13,42 @@ import (
...
 
-// Reset asserts the switch core reset field causing the SoC to restart with a
-// soft reset.
-//
-// Note that only the SoC itself is guaranteed to restart as, depending on the
-// board hardware layout, the system might remain powered (which might not be
-// desirable). See respective board packages for cold reset options.
-func Reset() {
+// ResetSystem initiates a system soft reset. The switch core always resets;
+// the VCore also resets unless protected through CPU_RESET_PROT_STAT.
```
abarisani: Simplify this description, make documentation clear to users and a reset should be a simple matter. For a user what does "protected through CPU_RESET_PROT_STAT" mean or entail?

abarisani (changes requested):

soc/microchip/lan969x/reset.go
```diff
@@ -13,16 +13,31 @@ import (
...
-// Reset asserts the switch core reset field causing the SoC to restart with a
-// soft reset.
-//
-// Note that only the SoC itself is guaranteed to restart as, depending on the
-// board hardware layout, the system might remain powered (which might not be
-// desirable). See respective board packages for cold reset options.
-func Reset() {
-	reg.Set(GCB_SOFT_RST, SOFT_SWC_RST)
+// ResetCPU asserts the selected CPU_RESET fields.
+func ResetCPU(mask uint32) {
```
abarisani: While the register belongs to CPU_REGS it's really a generic reset register. I would rename `ResetCPU` to `func Reset(target ResetTarget)`

`ResetTarget` is a `uint8` type with a iota that describes the possible settings with a readable variable, each with a description of the meaning. The `Reset` argument can then take multiple values ORed together.

abarisani: I'd use the same logic of the previous comment also for `GCP_SOFT_RST` and have a single `func SoftReset(prot ResetProtection)` function, the mask would be written to `CPU_RESET_PROT_STAT` and allow full control.

`ResetProtection` is a `uint8` type with a iota that describes the possible settings with a readable variable, each with a description of the meaning. The `SoftReset` argument can then take multiple values ORed together.

abarisani (changes requested):

soc/microchip/lan969x/reset.go
```diff
@@ -13,16 +13,42 @@ import (
...
-//
-// Note that only the SoC itself is guaranteed to restart as, depending on the
-// board hardware layout, the system might remain powered (which might not be
-// desirable). See respective board packages for cold reset options.
-func Reset() {
-	reg.Set(GCB_SOFT_RST, SOFT_SWC_RST)
+// ResetTarget specifies what Reset restarts.
+type ResetTarget uint8
+
+const (
```
abarisani, line 28: Map all 8 mask bits please.

soc/microchip/lan969x/reset.go
```diff
@@ -13,16 +13,42 @@ import (
...
+	// Core0WarmReset restarts the primary ARM core without a full processor reboot.
+	Core0WarmReset ResetTarget = 1 << (iota + 6)
+	// VCoreReset performs a full processor reboot without restarting the entire chip.
+	VCoreReset
+)
+
+// ResetProtection controls VCore propagation of a switch-core soft reset.
+type ResetProtection uint8
+
+const (
```
abarisani, line 51: Map all 6 bits please.

abarisani (changes requested):

PR #111, opened 2026-09-17
Add Spacemit K1 and BPI F3 support

micgor32:
Adds support for Spacemit K1 SoC and BPI F3 board. Marked as a draft as I want to add few more drivers (eth and trng for eg). The state for now is that the code was tested and boots on a real hw.


board/bananapi/f3/console.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/usbarmory/tamago/soc/spacemit/k1"
+)
+
+//go:linkname printk runtime/goos.Printk
+func printk(c byte) {
+	k1.UART0.TrySyncTx(c, 1<<22)
+
+	// The runtime terminates lines with a line feed alone, which advances
+	// a terminal to the next row without returning to the first column,
+	// staircasing the output.
```
MDr164, line 25: The staircasing is a global tamago issue and I so far fixed it locally in my actual `main()`, I'm not sure if board-level is the right place for it. My previous attempt at doing it board level was denied iirc.

board/bananapi/f3/mem.go
```diff
@@ -0,0 +1,22 @@
...
+//go:build !linkramsize
+
+package f3
+
+import (
+	_ "unsafe"
+)
+
+// The BPI-F3 is available with 2, 4, 8 or 16 GB of LPDDR4x DRAM, the default
+// size below is therefore conservative and covers all variants. Applications
```
MDr164, line 18: Out of curiosity: If the lowest size is 2G, why choose 512M as the default?

MDr164 (commented):
While I'm personally a fan of the larger explanatory comments I know that Andrea is not, I assume those need to be slimmed down to only the most crucial information.

board/bananapi/f3/mem.go
micgor32: no particular reason, was an arbitrary value that is large enough for testing

abarisani:
> While I'm personally a fan of the larger explanatory comments I know that Andrea is not, I assume those need to be slimmed down to only the most crucial information.

I can take a look and pinpoint my preference, maybe I can be persuaded :). 

Thanks for considering my preference.

